PR #59, opened 2020-08-04
Create localized instance of Translator in SitesGenerator.

tmeyer2115:
This PR has SitesGenerator creating a Translator instance per locale. This
instance is used to create the localized HBS helpers needed to build the locale's
pages.

A couple other small changes were included in this PR. The RepositoryScaffolder
had a bug where it generated a Jambo config with a dirs.translsations attribute.
This should be dirs.translations. The LocalFileParser was changed to gracefully
handle the case when translation files don't exist for a locale.

TEST=auto, manual

Added unit tests for the new LocalFileParser behavior. Locally, I created a Jambo
site with a French locale. I provided some French translations. One of my custom
partials used the new translate helper with interpolation. Confirmed that the
text was translated correctly. I also verified that if I tried translating a
string for which there were no translations, it simply returned the string.

I made sure that I could still build pages for a Jambo site that did not have
any local translations.

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -68,6 +68,10 @@ exports.SitesGenerator = class {
       }
     });
 
+    console.log('Registering Jambo Handlebars helpers');
+    // Register needed Handlebars helpers.
+    this._registerHelpers();
+
     console.log('Registering all handlebars templates');
     // Register Theme partials.
     const defaultTheme = config.defaultTheme;
@@ -93,7 +97,7 @@ exports.SitesGenerator = class {
     const translations = 
       config.dirs.translations ? await this._extractTranslations(locales) : {};
 
-    for (let locale of locales) {
+    for (const locale of locales) {
       console.log(`Writing files for '${locale}' locale`);
 
       const localeFallbacks = GENERATED_DATA.getLocaleFallbacks(locale);
@@ -107,9 +111,9 @@ exports.SitesGenerator = class {
       }).build();
 
       const translator = await Translator.create(locale, localeFallbacks, translations);
-      console.log('Registering Jambo Handlebars helpers');
-      // Register needed Handlebars helpers.
-      this._registerHelpers(translator);
+      console.log('Registering Jambo Handlebars translation helpers');
+      // Register needed Handlebars translation helpers.
+      this._registerTranslationHelpers(translator);
 
       new PageWriter({
         pagesDirectory: config.dirs.pages,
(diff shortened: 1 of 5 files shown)


src/commands/build/sitesgenerator.js
```diff
@@ -300,6 +309,24 @@ exports.SitesGenerator = class {
       return translator.translate(phrase, interpValues);
     });
   }
+  
+  /**
+   * Parses the local translation files for the provided locales. The translations
+   * are returned in i18next format.
+   * 
+   * @param {Array<string>} locales The list of locales.
```
oshi97, line 317: nit but can we have return type here

src/commands/build/sitesgenerator.js
```diff
@@ -91,17 +89,28 @@ exports.SitesGenerator = class {
     ];
     this._createStaticOutput(staticDirs, config.dirs.output);
 
-    for (let locale of GENERATED_DATA.getLocales()) {
+    const locales = GENERATED_DATA.getLocales();
+    const translations = 
+      config.dirs.translations ? await this._extractTranslations(locales) : {};
+
+    for (let locale of locales) {
```
oshi97, line 96: nit but can we use const here, and separate this looped logic into a method, maybe a `this._writePage()`?

src/commands/build/sitesgenerator.js
```diff
@@ -91,17 +89,28 @@ exports.SitesGenerator = class {
...
         locale: locale,
-        localeFallbacks: GENERATED_DATA.getLocaleFallbacks(locale),
+        localeFallbacks,
         urlFormatter: GENERATED_DATA.getUrlFormatter(locale),
       }).build();
 
+      const translator = await Translator.create(locale, localeFallbacks, translations);
+      console.log('Registering Jambo Handlebars helpers');
+      // Register needed Handlebars helpers.
+      this._registerHelpers(translator);
```
oshi97, line 112: maybe we could separate register helpers into regsisterTranslationHelpers and registerHelpers? It feels a little weird registering the same helpers over and over again.

src/commands/build/sitesgenerator.js
cea2aj: I'm all for putting this in a separate method

tmeyer2115: Changed the usage from let to const. If it's alright, I will leave the logic within the loop as is. Alexis is currently working on a big-ol refactor, so that should clean this up.

oshi97: ah sweet sounds good to me then!

PR #60, opened 2020-08-04
Rename directory to match its imports

alexisgrow:
The directory name didn't match its imports. Chose to change the directory name from `directAnswerCard` to `directanswercard` to match the the convention in this repo.

TEST=manual


alexisgrow:
Yes!

PR #64, opened 2020-08-05
Add the translatePlural helper to Jambo.

tmeyer2115:
This PR adds a translatePlural HBS helper to Jambo. This helper can provide
both singular and plural forms of translation. It also supports interpolation.
The helper uses a Translator instance under the hood. The Translator class
now has a translatePlural method.

J=SPR-463
TEST=auto, manual

Added new unit tests for Translator.translatePlural. Created a French Jambo
site locally. Added a custom partial that invoked the translatePlural
helper. Verified that singular and plural forms were translated properly. Also
verified that interpolation and translation fallbacks worked correctly with
the new helper.


src/commands/build/sitesgenerator.js
```diff
@@ -279,6 +279,18 @@ exports.SitesGenerator = class {
...
     });
+
+    /**
+     * Translates the provided phrase. The translation will be pluralized depending
+     * on the count. Interpolation is supported for both singular and plural forms.
+     */
+    hbs.registerHelper(
+      'translatePlural', 
+      function (singularPhrase, pluralPhrase, count, options) {
+        const interpValues = options.hash;
```
oshi97, line 290: I think this might force people to have count as the 3rd argument? Like I think the below wouldn't work, and they would have to switch count and cookieType
```hbs
{{ translatePlural
  'a {cookieType} cookie'
  '{count} {cookieType} cookies'
  cookieType='chocolate'
  count='3'
}}
```

tests/fixtures/translations/fr-FR.po
```diff
@@ -17,6 +17,11 @@ msgstr[1] "Articles"
 msgid "Hello {{name}}"
 msgstr "Bonjour {{name}}"
 
+msgid "There is {{count}} item {{name}}"
```
oshi97, line 20: we could do this in a separate pr too, but could we use the double braces to single braces, and also update the i18next.init to use single braces for interpolation.prefix and interpolation.suffix? Otherwise it doesn't work with handlebars. https://www.i18next.com/translation-function/interpolation#additional-options

tmeyer2115: I tested this syntax manually in Handlebars and it was working:

{ translatePlural "There is {{count}} item {{name}}" }

HBS seemed to ignore any double curly braces inside a string.

src/commands/build/sitesgenerator.js
tmeyer2115: Count does need to be the 3rd argument, yes. IMO, that's okay as long as we document it. I wanted to separate it from the variadic interpolation params. This, in my mind, designates that it is "special" and that it stands on its own, like the singular and plural forms.  I'm sure that these interfaces will be subject to some change as we start getting feedback from the HHs.

tests/fixtures/translations/fr-FR.po
oshi97: yup! pr here https://github.com/yext/jambo/pull/65 sorry for the fake news

src/commands/build/sitesgenerator.js
oshi97: Ah I see, sgtm!

PR #66, opened 2020-08-06
sitesgenerator: only print jambo injected data not entire process.env

oshi97:
This commit changes envvarparser to only parser out an env object
with the env keys that are asked for, instead of the entire environment.

TEST=manual

run jambo build before changes, see entire process.env printed
run jambo build after changes, see only JAMBO_INJECTED_DATA printed

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -30,7 +30,7 @@ exports.SitesGenerator = class {
     // Pull all data from environment variables.
     const envVarParser = EnvironmentVariableParser.create();
     const env = envVarParser.parse(['JAMBO_INJECTED_DATA'].concat(jsonEnvVars));
-    console.log('Jambo Injected Data:', env);
+    console.log('Jambo Injected Data:', env['JAMBO_INJECTED_DATA']);
 
     console.log('Reading config files');
     const pagesConfig = {};


src/commands/build/sitesgenerator.js
```diff
@@ -30,7 +30,7 @@ exports.SitesGenerator = class {
     // Pull all data from environment variables.
     const envVarParser = EnvironmentVariableParser.create();
     const env = envVarParser.parse(['JAMBO_INJECTED_DATA'].concat(jsonEnvVars));
-    console.log('Jambo Injected Data:', env);
+    console.log('Jambo Injected Data:', env['JAMBO_INJECTED_DATA']);
```
alexisgrow, line 33: Would it make sense to only have the "JAMBO_INJECTED_DATA" in env? All of this data is passed through to the template, it seems like there's a lot we don't need/want

bryanreed: I think we probably want to not prescribe that the theme uses this specific variable - given this is so urgent I think this is a good short term fix - I think what we probably want to do longer term is remove this printing and expose env vars as a first class thing in SGS

oshi97: Yeah I'm talking with tom about envVarParser not sure what we want it to do

PR #68, opened 2020-08-06
Add translateWithContext helper

cea2aj:
Adds a translateWithContext HBS helper to Jambo. Interpolation is also supported.
The Translator class now has a translateWithContext method.

J=SLAP-464
TEST=auto, manual

Added new unit tests for Translator.translateWithContext that include simple translations with context and interpolated translations with context.

I also tested a local french jambo site with a partial that invokes the helper

### DIFF
--- a/src/i18n/translator/translator.js
+++ b/src/i18n/translator/translator.js
@@ -49,8 +49,8 @@ class Translator {
   }
 
   /**
-     * Translates the provided phrase depending on the context.
-     * Supports interpolation.
+   * Translates the provided phrase depending on the context.
+   * Supports interpolation.
    * 
    * @param {string} phrase The phrase to translate.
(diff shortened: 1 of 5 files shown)


src/i18n/translator/translator.js
```diff
@@ -48,6 +48,21 @@ class Translator {
       this._i18next.t(pluralForm, parsedInterpValues);
   }
 
+  /**
+     * Translates the provided phrase depending on the context.
+     * Supports interpolation.
```
oshi97, line 53: I want to say this indent is weird but I honestly like it lol. Just pointing it out in case it was by accident

cea2aj: That was an accident, haha. But you're right, it kind of works

PR #69, opened 2020-08-11
build/sitesgenerator: Add babel plugins from SDK

creotutar:
We removed global polyfills in the SDK. The hbs JS was relying on these global
polyfills for browser support. We add the transforms from the SDK to the sitebuilder
so that they are transpiled to ES5. This is in addition to adding the transforms to 
the bundle JS in the theme.

J=SLAP-555
TEST=manual

Test basic functionality for a Jambo site referencing a local SDK dist (without
global polyfills) in Edge, IE11 on Windows. In Firefox, Chrome on MacOS.
On Safari on iOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,23 @@ The build command uses the 'defaultTheme' attribute in the jambo.json in Jambo's
 }
 ```
 
+##### Helpers
+
+The build command registers a few handlebars helpers for convenience:
+
+* `json(context)`: Renders the context as JSON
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -119,6 +119,23 @@ The build command uses the 'defaultTheme' attribute in the jambo.json in Jambo's
 }
 ```
 
+##### Helpers
```
alexisgrow, line 122: 😍

PR #73, opened 2020-08-13
Add unit tests for jambo build

alexisgrow:
TEST=unit,manual

### DIFF
--- a/tests/commands/build/globalconfiglocalizer.js
+++ b/tests/commands/build/globalconfiglocalizer.js
@@ -3,7 +3,7 @@ const GlobalConfigLocalizer = require('../../../src/commands/build/globalconfigl
 const LocalizationConfig = require('../../../src/models/localizationconfig');
 
 describe('GlobalConfigLocalizer works for sites without a localization config', () => {
-  it('creates localized global config when localization config is empty', () => {
+  it('does not alter global config when localization config is empty', () => {
     const globalConfig = new GlobalConfig({
       locale: 'es',
       apiKey: 'apiKey',
--- a/tests/commands/build/globalconfiglocalizer.js
+++ b/tests/commands/build/globalconfiglocalizer.js
@@ -1,0 +1,88 @@
+const GlobalConfig = require('../../../src/models/globalconfig');
+const GlobalConfigLocalizer = require('../../../src/commands/build/globalconfiglocalizer');
+const LocalizationConfig = require('../../../src/models/localizationconfig');
+
+describe('GlobalConfigLocalizer works for sites without a localization config', () => {
+  it('creates localized global config when localization config is empty', () => {
+    const globalConfig = new GlobalConfig({
+      locale: 'es',
+      apiKey: 'apiKey',
+      experienceKey: 'experienceKey',
+    });
+    const localizationConfig = new LocalizationConfig();
+    const localeToGlobalConfig = new GlobalConfigLocalizer(localizationConfig)
+      .localize(globalConfig, globalConfig.getLocale());
+
+    expect(localeToGlobalConfig).toEqual(globalConfig);
+  });
+});
+
+describe('GlobalConfigLocalizer generates expected localized global configs', () => {
+  const globalConfig = new GlobalConfig({
+    locale: 'es',
+    apiKey: 'apiKeyFromGlobalConfig',
+    experienceKey: 'experienceKeyFromGlobalConfig'
+  });
+
+  const localizationConfig = new LocalizationConfig({
+    localeConfig: {
+      en: {
+        apiKey: 'en_apiKeyFromLocalizationConfig',
+        experienceKey: 'en_experienceKeyFromLocalizationConfig'
+      },
+      es: {
+        apiKey: 'es_apiKeyFromLocalizationConfig',
+      },
+      de: {
+        experienceKey: 'de_experienceKeyFromLocalizationConfig'
+      },
+      fr: {}
+    }
+  });
+
+  it('uses localized values if provided in localization config', () => {
+    const localeToGlobalConfig = new GlobalConfigLocalizer(localizationConfig)
+      .localize(globalConfig, 'en');
+
+    expect(localeToGlobalConfig).toEqual(
+      new GlobalConfig({
+        locale: 'en',
+        apiKey: 'en_apiKeyFromLocalizationConfig',
+        experienceKey: 'en_experienceKeyFromLocalizationConfig'
+      }));
+  });
+
+  it('merges global config and localization config', () => {
+    const localeToGlobalConfig = new GlobalConfigLocalizer(localizationConfig)
+      .localize(globalConfig, 'es');
+
+    expect(localeToGlobalConfig).toEqual(
+      new GlobalConfig({
+        locale: 'es',
+        apiKey: 'es_apiKeyFromLocalizationConfig',
+        experienceKey: 'experienceKeyFromGlobalConfig'
+      }));
+
+    const localeToGlobalConfig2 = new GlobalConfigLocalizer(localizationConfig)
+      .localize(globalConfig, 'de');
+
+    expect(localeToGlobalConfig2).toEqual(
+      new GlobalConfig({
+        locale: 'de',
+        apiKey: 'apiKeyFromGlobalConfig',
+        experienceKey: 'de_experienceKeyFromLocalizationConfig'
+      }));
+  });
+
+  it('uses values from global config if absent in localization config', () => {
+    const localeToGlobalConfig = new GlobalConfigLocalizer(localizationConfig)
+      .localize(globalConfig, 'fr');
+
+    expect(localeToGlobalConfig).toEqual(
+      new GlobalConfig({
+        locale: 'fr',
+        apiKey: 'apiKeyFromGlobalConfig',
+        experienceKey: 'experienceKeyFromGlobalConfig'
+      }));
+  });
+});
--- a/tests/commands/build/globalconfiglocalizer.js
+++ b/tests/commands/build/globalconfiglocalizer.js
@@ -23,62 +23,27 @@ describe('GlobalConfigLocalizer generates expected localized global configs', ()
     apiKey: 'apiKeyFromGlobalConfig',
     experienceKey: 'experienceKeyFromGlobalConfig'
   });
-
   const localizationConfig = new LocalizationConfig({
     localeConfig: {
       en: {
         apiKey: 'en_apiKeyFromLocalizationConfig',
         experienceKey: 'en_experienceKeyFromLocalizationConfig'
       },
-      es: {
-        apiKey: 'es_apiKeyFromLocalizationConfig',
-      },
-      de: {
-        experienceKey: 'de_experienceKeyFromLocalizationConfig'
-      },
       fr: {}
     }
   });
 
-  it('uses localized values if provided in localization config', () => {
-    const localeToGlobalConfig = new GlobalConfigLocalizer(localizationConfig)
-      .localize(globalConfig, 'en');
+  it('merges global config and localization config correctly', () => {
+    const globalConfigLocalizer = new GlobalConfigLocalizer(localizationConfig);
 
-    expect(localeToGlobalConfig).toEqual(
+    expect(globalConfigLocalizer.localize(globalConfig, 'en')).toEqual(
       new GlobalConfig({
         locale: 'en',
         apiKey: 'en_apiKeyFromLocalizationConfig',
         experienceKey: 'en_experienceKeyFromLocalizationConfig'
       }));
-  });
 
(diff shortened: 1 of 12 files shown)


tests/commands/build/globalconfiglocalizer.js
```diff
@@ -0,0 +1,88 @@
...
+
+    expect(localeToGlobalConfig).toEqual(
+      new GlobalConfig({
+        locale: 'en',
+        apiKey: 'en_apiKeyFromLocalizationConfig',
+        experienceKey: 'en_experienceKeyFromLocalizationConfig'
+      }));
+  });
+
+  it('merges global config and localization config', () => {
```
tmeyer2115, line 55: I think we can consolidate these tests a bit. Instead of having three different tests for these things, we could have a single test: "merges global config and localization config correctly". We can construct test fixtures such that we're verifying that localizaiton config takes precedence, but global config values are used if not present there. 

Unit tests are a bit of a balancing act. Too many tests and our suite becomes slow to run. Too few and we don't have total coverage. If a class has a lot going on, it may be better to err on the side of more unit tests (that are small and highly targeted). That way, when a test(s) fail, we can more easily pinpoint what about that class is broken. This particular class, though, is pretty straightforward. If out single test fails here, we can easily compare the output and the code to see what's wrong.

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
+const PageConfig = require('../../../src/models/pageconfig');
+const PageConfigDecorator = require('../../../src/commands/build/pageconfigdecorator');
+
+describe('PageConfigDecorator decorates PageConfigs and builds the expected object', () => {
+  it('builds locale to configs with decorated configs when no locales specified', () => {
```
tmeyer2115, line 5: "when only default locale is specified"? "when no locales specified" makes me think there should be no map

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
...
+        new PageConfig({
+          locale: defaultLocale,
+          pageName: 'pageName2',
+          rawConfig: { pageTitle: 'pageTitle' },
+        })
+      ]
+    });
+  });
+
+  it('builds locale to configs with decorated configs with given locales', () => {
```
tmeyer2115, line 37: 'builds decorated page configs for specified locales'? That's a bit more streamlined. It also gets at the point of the PageConfigDecorator class without leaking implementation details (that it's a map).

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
...
+        new PageConfig({
+          pageName: 'example',
+          locale: 'fr',
+          rawConfig: { verticalKey: 'FR' },
+        }),
+      ],
+    });
+  });
+
+  it('decorated config props override regular config props', () => {
```
tmeyer2115, line 72: I'm a bit confused by this test name. I'm not sure what's meant by 'regular config props'. This makes it sound a bit like the PageConfigDecorator favors a fallback locale's configuration vs the actual locale's config.

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
...
+    new PageConfig({
+      pageName: 'example',
+      locale: 'en',
+      rawConfig: {
+        example: 'ex'
+      },
+    }),
+  ];
+
+  it('decorating config with fallback and default config works', () => {
```
tmeyer2115, line 188: I'm wondering here too if we can consolidate the tests in this describe by having more inclusive fixtures. Jest is nice because it highlights the differences between expected v. actual. This helps us pinpoint what exactly may be broken in the class without needing separate tests for the different bits of functionality. 

Broadly, we should write just enough tests so that we can have full coverage and be able to pinpoint what's causing a test to fail with relative ease.

tests/commands/build/templatedirector.js
```diff
@@ -0,0 +1,168 @@
+const TemplateDirector = require('../../../src/commands/build/templatedirector');
+const PageTemplate = require('../../../src/models/pagetemplate');
+
+describe('TemplateDirector directs PageTemplates and builds the expected object', () => {
+  it('creates localeToPageTemplates properly when locale data is absent', () => {
```
tmeyer2115, line 5: I think we want to reword the test description. Someone reading this would expect that something called localeToPageTemplates is empty when there is no locale data. "creates page templates correctly when only defaultLocale is present"?

tests/commands/build/templatedirector.js
```diff
@@ -0,0 +1,168 @@
...
+          pageName: 'path',
+          locale: 'fr',
+          path: `pages/path.fr.html.hbs`
+        }),
+      ],
+    });
+  });
+});
+
+describe('Finds the correct PageTemplate given locale information', () => {
```
tmeyer2115, line 109: Is the test above more general and these ones more targeted?

tests/commands/build/templatedirector.js
```diff
@@ -0,0 +1,168 @@
...
+
+    const template = new TemplateDirector({
+      locales: locales,
+      localeToFallbacks: localeToFallbacks
+    })._findPageTemplateForLocale(locale, pageTemplates);
+
+    expect(template).toEqual(localeToPageTemplate[locale]);
+  });
+
+  it('finds template correct fallback locale', () => {
```
tmeyer2115, line 143: I think there might be a typo here

tests/models/configurationregistry.js
```diff
@@ -0,0 +1,83 @@
...
+          pageName: getPageName(configNameWithLocale),
+          locale: ConfigurationRegistry._parseLocale(configNameWithLocale),
+          rawConfig: rawPageConfig
+        })
+      ]);
+    expect(rawConfigs).toEqual(rawConfigsCopy);
+  });
+});
+
+describe('ConfigurationRegistry parses locale from config name', () => {
```
tmeyer2115, line 74: I think we can get away without these. The tests you have above would break if _parseLocale was broken and it seems like it'd be quick to tell from those failures that _parseLocale was the culprit.

tests/models/generateddata.js
```diff
@@ -0,0 +1,327 @@
...
+        'de': {},
+      }
+    });
+
+    const locales = new GeneratedData({
+      defaultLocale: 'fr',
+      localizationConfig: localizationConfig
+    }).getLocales();
+
+    return expect(locales).toEqual(localizationConfig.getLocales());
```
tmeyer2115, line 24: I'm a bit confused by this test. defaultLocale should be something that's in the localeConfig (I think). They can't specify a defaultLocale which has no locale-specific config in locale_config.

tests/models/localizationconfig.js
```diff
@@ -0,0 +1,109 @@
...
+describe('Getting URL Formatting function works properly', () => {
+  it('falls back to "pageName.pageExt" when no localization config provided', () => {
+    let locale = 'en';
+    let localizationConfig = new LocalizationConfig();
+
+    let urlFormatter = localizationConfig.getUrlFormatter(locale);
+    expect(urlFormatter('pageName', 'pageExt')).toEqual('pageName.pageExt');
+  });
+
+  it('formats with default url formatting pattern', () => {
```
tmeyer2115, line 48: 'formats defaultLocale with correct url formatting pattern'? In this test and the two below, we want the test name to make clear not just that we're both selecting the correct formatting function and that that formatting function is working properly.

tests/commands/build/globalconfiglocalizer.js
```diff
@@ -0,0 +1,88 @@
+const GlobalConfig = require('../../../src/models/globalconfig');
+const GlobalConfigLocalizer = require('../../../src/commands/build/globalconfiglocalizer');
+const LocalizationConfig = require('../../../src/models/localizationconfig');
+
+describe('GlobalConfigLocalizer works for sites without a localization config', () => {
+  it('creates localized global config when localization config is empty', () => {
```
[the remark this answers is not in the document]
alexisgrow: Yes totally, what do you think of: "does not alter global config when localization config is empty"?

tests/commands/build/globalconfiglocalizer.js
alexisgrow: That makes sense! Reduced this to one test!

tests/models/generateddata.js
alexisgrow: Ah fair, so for backwards compatibility, if there is no locale_config.json file, it uses the locale specified in the global_config file for the "defaultLocale" (and if that doesn't exist, it uses the empty string ("") for the default locale). This allows us to treat sites that are single-locale or multi-locale similarly!

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
+const PageConfig = require('../../../src/models/pageconfig');
+const PageConfigDecorator = require('../../../src/commands/build/pageconfigdecorator');
+
+describe('PageConfigDecorator decorates PageConfigs and builds the expected object', () => {
+  it('builds locale to configs with decorated configs when only default locale is specified', () => {
```
tmeyer2115, line 5: 'builds decorated pages configs correctly when there is only a default locale'?

tests/commands/build/pageconfigdecorator.js
alexisgrow: Consolidated to one test here!

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
...
+        new PageConfig({
+          pageName: 'example',
+          locale: 'fr',
+          rawConfig: { verticalKey: 'FR' },
+        }),
+      ],
+    });
+  });
+
+  it('locale-specific config props override fallback and default config props', () => {
```
tmeyer2115, line 72: I'm not sure I follow this test description. The descriptions should read like normal sentences (or at least fragments of one). I try and base my test descriptions on how I would verbally describe the test to another engineer.

tests/commands/build/pageconfigdecorator.js
```diff
@@ -0,0 +1,273 @@
...
+              verticalKey: 'default'
+            },
+          },
+        }),
+      ],
+    });
+  });
+});
+
+describe('Decorating a single PageConfig works properly', () => {
```
tmeyer2115, line 137: The tests above also seem to ensure that we're decorating PageConfigs properly. Are these tests at all redundant?

tests/models/localizationconfig.js
```diff
@@ -0,0 +1,109 @@
...
+        baseLocale: 'pages/{pageName}.{pageExt}',
+        default: 'pages/{locale}/{pageName}.{pageExt}'
+      }
+    });
+
+    let urlFormatter = localizationConfig.getUrlFormatter(locale);
+    expect(urlFormatter('pageName', 'pageExt')).toEqual('pages/pageName.pageExt');
+  });
+
+  it('formats with baseLocale url formatting pattern', () => {
```
tmeyer2115, line 62: 'applies correct url formatting pattern to non-default locales'?

tests/commands/build/templatedirector.js
alexisgrow: Removed as part of simplification of tests!

tests/commands/build/templatedirector.js
alexisgrow: I updated the fixture to become more complicated, so one test above has coverage for these cases.

tests/models/localizationconfig.js
```diff
@@ -0,0 +1,109 @@
...
+        default: 'pages/{locale}/{pageName}.{pageExt}'
+      }
+    });
+
+    let locale = 'en';
+    let urlFormatter = localizationConfig.getUrlFormatter(locale);
+    expect(urlFormatter('pageName', 'pageExt')).toEqual(`pages/${locale}/pageName.pageExt`);
+  });
+
+  it('formats with overrides for a given locale', () => {
```
tmeyer2115, line 76: 'applies correct url formatting pattern for locale with urlOverride'? As I mentioned previously, we want to highlight that these tests cover two things: we're selecting the right formatter and that formatter is constructed properly.

tests/commands/build/pageconfigdecorator.js
alexisgrow: Deleted this test and made a new "builds decorated pages configs correctly when there are multiple locales and fallbacks" test!

tests/commands/build/pageconfigdecorator.js
alexisgrow: Consolidated these cases into a single tests called "builds decorated pages configs correctly when there are multiple locales and fallbacks"

alexisgrow:
🎊🎊🎊🎊🎊
 
unit_tests

PR #74, opened 2020-08-13
Add a preprocessor for all partials registered with Handlebars.

tmeyer2115:
This PR adds the PartialPreprocessor class. This class performs some processing
on partials before Jambo registers them with Handlebars. Specifically, it looks
to transpile all usages of the 'translate' helper from a partial. If the
translation can be resolved at compile-time, the 'translate' call is replaced
with the translated string. If the translation must be done at run-time, the
call is replaced with a call to the SDK's new run-time translation methods.
Translations must be done at run-time if they require pluralization or
interpolation.

Currently, the transpilation only works for JS partials. A subsequent PR will
add support for normal .hbs partials as well.

The Translator class will need to be updated to support run-time translations.
Specifically, the translatePlural method will provide all singular and plural
forms of the string. The SDK's run-time method can then choose the right form
at run-time. The translate and translateWithContext methods will need to be
updated as well to supply a formatter string when interpolation is required.
The SDK's run-time method will take the formatter string and the interpolation
options and produce the right translation.

J=SLAP-557
TEST=auto

Added unit tests for the PartialPreprocessor. In the tests, I mocked the
Translator class to have the behavior outlined above.

### DIFF
--- a/src/partials/partialpreprocessor.js
+++ b/src/partials/partialpreprocessor.js
@@ -11,9 +11,9 @@ class PartialPreprocessor {
   }
 
   /**
-   * Processes the provided partial, looking for usages of the 'translate'
-   * helper. These usages are transpiled into either a translated string or a call
-   * to the SDK's run-time translation methods.
+   * Processes the provided partial, looking for usages of the 'translate' or
+   * 'translateJS' helpers. These usages are transpiled into either a translated
+   * string or a call to the SDK's run-time translation methods.
    * 
    * @param {string} partial The partial to process.
    * @returns {string} The transpiled partial.
@@ -21,7 +21,7 @@ class PartialPreprocessor {
   process(partial) {
     let processedPartial = partial;
     const translateHelperCalls =
-      processedPartial.match(/\{\{\s?translate\s.+\}\}/g) || [];
+      processedPartial.match(/\{\{\s?translate(JS)?\s.+\}\}/g) || [];
 
     translateHelperCalls.forEach(call => {
       const translateInvocation = TranslateInvocation.from(call);
@@ -34,9 +34,9 @@ class PartialPreprocessor {
   }
 
   /**
-   * Transpiles a usage of the 'translate' helper. If the translation
-   * can be resolved at compile-time (no pluralization or interpolation), the
-   * usage will be transpiled to it. Otherwise, it will be transpiled to the
+   * Transpiles a usage of the 'translate' or 'translateJS' helper. If the 
+   * translation can be resolved at compile-time (no pluralization or interpolation), 
+   * the usage will be transpiled to it. Otherwise, it will be transpiled to the
    * appropriate call to the SDK's run-time translation method.
    * 
    * @param {TranslateInvocation} invocation The {@link TranslateInvocation} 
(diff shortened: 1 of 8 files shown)


src/partials/partialpreprocessor.js
```diff
@@ -0,0 +1,67 @@
...
+   * helper. These usages are transpiled into either a translated string or a call
+   * to the SDK's run-time translation methods.
+   * 
+   * @param {string} partial The partial to process.
+   * @returns {string} The transpiled partial.
+   */
+  process(partial) {
+    let processedPartial = partial;
+    const translateHelperCalls =
+      processedPartial.match(/\{\{\stranslate\s.+\}\}/g) || [];
```
bryanreed, line 24: what about the translateJS helper?  Might as well do both and add this to the TranslateInvocation class

src/partials/partialpreprocessor.js
```diff
@@ -0,0 +1,67 @@
...
+          invocation.getPhrase(), translationContext) :
+        this._translator.translate(invocation.getPhrase);
+      translatorResult = `'${translatorResult}'`;
+    }
+    let interpolationParams = JSON.stringify(invocation.getInterpolationParams());
+    interpolationParams = interpolationParams.replace(/[\'\"]/g, '');
+
+    return invocation.canBeTranslatedStatically() ?
+        translatorResult :
+        `ANSWERS.translateJS(${translatorResult}, ${interpolationParams})`;
```
bryanreed, line 64: it seems as though we aren't considering whether or not the original helper call was translate or translateJS - if they call translate with plurals we don't want to be executing a translate javascript call, we only want to do this when they to translateJS, otherwise we need to call the SDK handlebars helper.  I think we should do all of the logic in here though since its just one more boolean getter you need to add to the TranslationInvocation object as I mentioned in a comment above

tests/partials/partialpreprocessor.js
```diff
@@ -0,0 +1,32 @@
+const path = require('path');
+const { readFileSync } = require('file-system');
+
+const Translator = require('../../src/i18n/translator/translator');
+const PartialPreprocessor = require('../../src/partials/partialpreprocessor');
+jest.mock('../../src/i18n/translator/translator')
+
+describe('PartialPreprocessor works correctly', () => {
```
bryanreed, line 8: we also want a test to make sure the processor works on .hbs syntaxed files (since there is a greater chance catching something wrong with our regexes when there are different character patterns in the file, and hbs markup files are the other use case)

src/partials/partialpreprocessor.js
```diff
@@ -0,0 +1,106 @@
...
+   * @param {Object|string} translatorResult The translation(s) for the phrase.
+   * @param {Object<string, ?>} interpolationParams The needed interpolation parameters
+   *                                                (including 'count').
+   * @returns {string} The string-ified call to ANSWERS.translateJS.
+   */
+  _createRuntimeCallForJS(translatorResult, interpolationParams) {
+    let parsedParams = JSON.stringify(interpolationParams);
+    parsedParams = parsedParams.replace(/[\'\"]/g, '');
+
+    return `ANSWERS.translateJS(${translatorResult}, ${parsedParams})`;
```
bryanreed, line 84: do you think it makes more sense for this method call to have the count parameter passed via top level (optionally) in addition to the nested variable object?  Since it needs to be treated specially?  More thinking how it would feel to use this function manually in the SDK - the nested object of params makes sense for the interpolation stuff but count feels special to me.  If so it would need to be passed to both as one could use the count number in the interpolation as well.

tmeyer2115: Sure, we could have the signature be `translateJS(translations, interpParams, count)`. You're right, count is special, it's not like the other interpolation parameters.

PR #76, opened 2020-08-13
Return proper error codes

cea2aj:
Jambo now returns the following codes:
| Exit code | Reason              |
|-----------|---------------------|
| 0         | Success             |
| 1         | Unhandled Exception |
| 13        | UserError           |
| 14        | SystemError         |

This commit adds an exitWithError() function to errorutils which will print the error and then terminate jambo. If a UserError or SystemError is passed to exitWithError, then jambo will exit with codes 13 or 14 respectively. Any other Errors passed will terminate jambo with exit code 1. I chose codes 13 and 14 since node process has definitions for most codes between 2 and 12.

In general, a UserError or a SystemError should be thrown so that they can be caught inside src/cli.js and passed to exitWithError().

J=SLAP-538
TEST=manual

I tested various fail cases and observed the exit codes:
| Exit Code | Error                                                   |
|-----------|---------------------------------------------------------|
| 1         | Invalid command                                     |
| 13        | Attempt to import invalid theme                         |
| 13        | Overriding without having the theme imported            |
| 13        | Attempt to import before running init                   |
| 13        | Add card with invalid card to fork                      |
| 13        | Create direct answer card in folder that already exists |
| 13        | Jambo config incorrectly named                          |
| 13        | Invalid jambo config                               |
| 13        | Global config incorrectly named                         |
| 13        | Invalid global config                                   |
| 13         | Override a path that doesn't exist                      |
| 13         | Add page with invalid template                          |
| 13         | Handlebars build syntax error                                   |
| 14        | Cannot find module fs-extra when upgrading theme        |
|14        | Attempting to upgrade when the site is not a git repo   |

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -11,6 +11,8 @@ const { ThemeUpgrader } = require('./commands/upgrade/themeupgrader');
 const { parseJamboConfig } = require('./utils/jamboconfigutils');
 const yargs = require('yargs');
 const fs = require('file-system');
+const SystemError = require('./errors/systemerror');
+const { exitWithError } = require('./utils/errorutils');
 
 const jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
 
@@ -34,7 +36,7 @@ const options = yargs
     argv => {
       const repositorySettings = new initCommand.RepositorySettings(argv);
       const repositoryScaffolder = new initCommand.RepositoryScaffolder();
-      repositoryScaffolder.create(repositorySettings).catch(console.log);
+      repositoryScaffolder.create(repositorySettings)
     })
   .command(
     'import',
@@ -48,7 +50,7 @@ const options = yargs
     },
     argv => {
       const themeImporter = new themeCommand.ThemeImporter(jamboConfig);
-      themeImporter.import(argv.theme, argv.addAsSubmodule).then(console.log).catch(console.log);
+      themeImporter.import(argv.theme, argv.addAsSubmodule).then(console.log);
     })
   .command(
     'override',
@@ -134,9 +136,11 @@ const options = yargs
     argv => {
       const themeUpgrader = new ThemeUpgrader(jamboConfig);
       themeUpgrader
-        .upgrade(jamboConfig.defaultTheme, argv.disableScript, argv.isLegacy)
-        .catch(console.error);
+        .upgrade(jamboConfig.defaultTheme, argv.disableScript, argv.isLegacy).catch((error) => {
+            exitWithError(new SystemError(error.message, error.stack));
+        });
     })
+  .strict()
   .argv;
 
   /**
--- a/src/cli.js
+++ b/src/cli.js
@@ -172,11 +172,11 @@ const options = yargs
     argv => {
       const themeUpgrader = new ThemeUpgrader(jamboConfig);
       themeUpgrader
-        .upgrade(jamboConfig.defaultTheme, argv.disableScript, argv.isLegacy).catch((e) => {
+        .upgrade(jamboConfig.defaultTheme, argv.disableScript, argv.isLegacy).catch((err) => {
           if(isCustomError(err)){
             exitWithError(err);
           }
-          exitWithError(new SystemError(error.message, error.stack));
+          exitWithError(new SystemError(err.message, err.stack));
         });
     })
   .strict()
--- a/src/cli.js
+++ b/src/cli.js
@@ -13,9 +13,16 @@ const yargs = require('yargs');
 const fs = require('file-system');
 const SystemError = require('./errors/systemerror');
 const UserError = require('./errors/usererror');
-const { exitWithError } = require('./utils/errorutils');
+const { exitWithError, isCustomError } = require('./utils/errorutils');
 
-const jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
+let jamboConfig;
+
+try{
+  jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
+} catch (e) {
+  exitWithError(e);
+}
+
 
 const options = yargs
 	.usage('Usage: $0 <cmd> <operation> [options]')
@@ -37,7 +44,7 @@ const options = yargs
     argv => {
       const repositorySettings = new initCommand.RepositorySettings(argv);
       const repositoryScaffolder = new initCommand.RepositoryScaffolder();
-      repositoryScaffolder.create(repositorySettings)
+      repositoryScaffolder.create(repositorySettings).catch((e) => {exitWithError(e)});
     })
   .command(
     'import',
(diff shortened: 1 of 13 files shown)


src/utils/errorutils.js
```diff
@@ -0,0 +1,16 @@
+const fs = require('fs');
+
+/**
+ * Print the error, and then forcefully end the 
+ * process with the specified exit code. Pending
+ * async operations will be lost.
+ * @param {Error} error
+ */
+exports.exitWithError = (err) => {
```
alexisgrow, line 9: Instead of using this `exitWithError` helper/util, what do you think about using `throw`? (e.g. `throw new UserException(...)`)? It would allow the caller to prevent immediately exiting if they want to catch any errors.

src/commands/upgrade/themeupgrader.js
```diff
@@ -29,7 +31,7 @@ exports.ThemeUpgrader = class {
   async upgrade(themeName, disableScript, isLegacy) {
     const themePath = path.join(this._themesDir, themeName);
     if (!fs.existsSync(themePath)) {
-      throw new Error(`theme "${themeName}" not found within the "${this._themesDir}" folder`)
+      exitWithError(new SystemError(`Theme "${themeName}" not found within the "${this._themesDir}" folder`));
```
alexisgrow, line 34: I think this may be a `UserError`, is this the case where the jambo.json's "defaultTheme" is not present in the jambo.json's "config.dirs.themes" directory? If so, we'd advise the user to run the jambo's import theme command.

src/utils/errorutils.js
cea2aj: I like the idea of using throw. I think it's a little cleaner. There are a couple downsides however: Every caller has to either not catch Errors at all, or check the type of caught Errors and throw them if they are a SystemError or a UserError. Using exitWithError ensures that jambo will actually exit. If we simply throw everything there is a risk that the caller catches it, logs it, and then jambo wouldn't exit correctly. Also, using throws in async code doesn't get caught by the caller unless its using the async/await syntax. Also, exitWithError ensures that it exits immediately. If we use throws then other async code can get run before it exits. Ultimately I think that exitWithError is simpler and less error prone, especially in async code, but let me know your thoughts!

alexisgrow: Internal helpers being so prescriptive about when the entire process exits is a little tricky, there may be cases where we don't want to kill the Jambo process on every error. Using `throw`s internally give us the flexibility to retry, cleanup, or otherwise gracefully handle the failure before exiting. This is especially true for async code when dealing with external resources that need to be closed. Each Jambo command could totally have a top-level catch statement in `cli.js` that could use this logic to `exitWithError`, but I'm thinking for many of the usages of this helper it makes more sense to pass the errors up.

cea2aj: I like that idea. I'll refactor

src/commands/upgrade/themeupgrader.js
cea2aj: Agreed!

src/cli.js
```diff
@@ -11,8 +11,18 @@ const { ThemeUpgrader } = require('./commands/upgrade/themeupgrader');
 const { parseJamboConfig } = require('./utils/jamboconfigutils');
 const yargs = require('yargs');
 const fs = require('file-system');
+const SystemError = require('./errors/systemerror');
+const UserError = require('./errors/usererror');
+const { exitWithError, isCustomError } = require('./utils/errorutils');
+
+let jamboConfig;
+
+try{
```
alexisgrow, line 20: Minor styling comment across the PR, in our repos we add a space before curly braces, e.g. `try {` and `} else {`; and a space before parenthesis in conditionals, e.g. `if (...) {`.

src/cli.js
```diff
@@ -34,7 +44,7 @@ const options = yargs
     argv => {
       const repositorySettings = new initCommand.RepositorySettings(argv);
       const repositoryScaffolder = new initCommand.RepositoryScaffolder();
-      repositoryScaffolder.create(repositorySettings).catch(console.log);
+      repositoryScaffolder.create(repositorySettings).catch((e) => {exitWithError(e)});
```
alexisgrow, line 47: styling nit: If the braces around `exitWithError(e)` aren't necessary here, would it be possible to remove?

src/cli.js
```diff
@@ -47,8 +57,12 @@ const options = yargs
           { description: 'import the theme as a submodule', default: true, type: 'boolean' });
     },
     argv => {
-      const themeImporter = new themeCommand.ThemeImporter(jamboConfig);
-      themeImporter.import(argv.theme, argv.addAsSubmodule).then(console.log).catch(console.log);
+      try{
+        const themeImporter = new themeCommand.ThemeImporter(jamboConfig);
+        themeImporter.import(argv.theme, argv.addAsSubmodule).then(console.log).catch((e) => {exitWithError(e)});
```
alexisgrow, line 62: styling nit: The line length here is long (it was before this change, but now that it's indented farther it draws my eye to it more), would it be possible to put the .then() and .catch() on new lines? (Similar comment about L175)

src/commands/build/sitesgenerator.js
```diff
@@ -39,33 +41,40 @@ exports.SitesGenerator = class {
...
         try {
           pagesConfig[pageId] = parse(fs.readFileSync(path, 'utf8'), null, true);
-        } catch (e) {
-          if (e instanceof SyntaxError) {
-            throw new Error('JSON SyntaxError: could not parse ' + path);
-          } else {
-            throw e;
-          }
+        } catch (err) {
+          throw new UserError(err.message, err.stack);
```
alexisgrow, line 45: Is it intentional to remove the "JSON SyntaxError: could not parse ${path}" message here? I think that was added a while back to help users find the erroring file.

src/cli.js
cea2aj: Fixed. Do you think we should adopt semistandard style like the SDK?

src/cli.js
cea2aj: yep!

src/cli.js
cea2aj: sure

src/commands/build/sitesgenerator.js
cea2aj: Added back

src/cli.js
```diff
@@ -47,8 +57,14 @@ const options = yargs
           { description: 'import the theme as a submodule', default: true, type: 'boolean' });
     },
     argv => {
-      const themeImporter = new themeCommand.ThemeImporter(jamboConfig);
-      themeImporter.import(argv.theme, argv.addAsSubmodule).then(console.log).catch(console.log);
+      try{
```
alexisgrow, line 60: For this and the other "try"s in this PR, there should be a space before the curly brace, e.g. `try {`

src/cli.js
```diff
@@ -135,8 +175,14 @@ const options = yargs
       const themeUpgrader = new ThemeUpgrader(jamboConfig);
       themeUpgrader
         .upgrade(jamboConfig.defaultTheme, argv.disableScript, argv.isLegacy)
-        .catch(console.error);
+          .catch(err => {
+            if(isCustomError(err)){
```
alexisgrow, line 179: For this `if`, there should be spaces between the curly brace and the parenthesis, for example: `if (isCustomError(err)) {`

PR #77, opened 2020-08-17
Remove unused translation Handlebars helpers

tmeyer2115:
The new PartialPreprocessor class makes it such that all translation happens either before HBS compilation or after, during run-time. The 'translate', 'translatePlural', and 'translateWithContext' helpers provided by Jambo have been made redundant with this addition and should no longer be used.

TEST=none

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -101,14 +101,6 @@ exports.SitesGenerator = class {
     const translations =
       config.dirs.translations ? await this._extractTranslations(locales) : {};
 
-    for (const locale of locales) {
-      const localeFallbacks = GENERATED_DATA.getLocaleFallbacks(locale);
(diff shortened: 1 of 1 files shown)


src/commands/build/sitesgenerator.js
```diff
@@ -101,14 +101,6 @@ exports.SitesGenerator = class {
     const translations =
       config.dirs.translations ? await this._extractTranslations(locales) : {};
```
cea2aj, line 102: This declaration is unused. Can we remove it?

tmeyer2115: I'd prefer to keep the `translations` and `_extractTranslations` here for now. In short order, we will be adding the `PartialPreprocessor` to the `SitesGenreator`. The preprocessor will require a `Translator` instance be created (using `translations`). I can definitely remove the `Translator` import though, that'll be quick to add back in when we need it.

cea2aj: Makes sense!

PR #82, opened 2020-08-18
Update PageTemplates to use a Partials domain model

alexisgrow:
We need to preprocess all the partials for internationalization work. This means we can't read in the file right before we write it, so instead of storing the templatePath on the page it makes more sense to store the contents of that file. This makes it simpler to preprocess or transpile the contents of the template file. 

TEST=unit,manual

Run `jambo build && grunt webpack` on local internationalized test site. View pages in browser and confirm they look as expected.

### DIFF
--- a/src/models/pagetemplate.js
+++ b/src/models/pagetemplate.js
@@ -50,6 +50,29 @@ module.exports = class PageTemplate {
   }
 
   /**
+   * Updates the locale
+   *
+   * @param {String} locale
+   */
+  setLocale(locale) {
+    this.locale = locale;
+    return this;
+  }
+
+  /**
+   * Creates a a copy of this PageTemplate
+   *
+   * @returns {PageTemplate}
+   */
+  clone () {
+    return new PageTemplate({
+      locale: this.locale,
+      path: this.path,
+      pageName: this.pageName
+    });
+  }
+
+  /**
    * Creates a @type {PageTemplate} from a given filename and path
    *
    * @param {String} filename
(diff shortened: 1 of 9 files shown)


src/models/pagetemplate.js
```diff
@@ -32,37 +30,54 @@ module.exports = class PageTemplate {
...
    * @returns {PageTemplate}
    */
-  static from (filename, path) {
+  static from (filename, path, fileContents) {
     if (!filename) {
       throw new Error('Error: no filename provided for page template');
     }
 
     return new PageTemplate({
       path: path,
```
tmeyer2115, line 79: Do we still need both the path and the contents? It feels a bit odd having both as what's in the file path may be different than the contents. If it's difficult to just use file contents, no worries. I don't want us to spend a ton of time hooking this stuff up.

src/models/partial.js
```diff
@@ -0,0 +1,49 @@
...
+  getName() {
+    return stripExtension(this.path);
+  }
+
+  /**
+   * Returns the partial's path
+   *
+   * @returns {String}
+   */
+  getPath() {
```
tmeyer2115, line 37: Same question about needing both the path and the contents

src/models/pagetemplate.js
alexisgrow: Yeah, that's a good point. I originally was thinking we should add the contents here because we're preprocessing the file contents, so we intend for the page template file contents to be different than what's in the file path! We use the path in the naming scheme when we register partials with the handlebars renderer as well as the name of the output file. Do you think we should refactor to store less information here?

PR #83, opened 2020-08-18
Add partial preprocessor to jambo build

alexisgrow:
TEST=unit,manual

Use the following helpers in a page.html.hbs file, run `jambo build` and see the output 
| Input | Output |
| :---: | :---: | 
| {{ translateJS phrase='Hello' }} | 'Hello' | 
| {{ translate phrase='Hello'}} | 'Hello' | 
| {{ translateJS phrase='Some item [[name]]' pluralForm='Some items [[name]]' name=profile.name count=profile.count }}| ANSWERS.translateJS('{"1":"Some item [[name]]"}', {name:profile.name,count:profile.count}, profile.count) | 
| {{ translateJS phrase='Mail now [[id1]]' context='Mail is a verb' id1=profile.name}} | ANSWERS.translateJS('Mail now [[id1]]', {id1:profile.name}) |


src/commands/build/sitesgenerator.js
```diff
@@ -97,12 +95,35 @@ exports.SitesGenerator = class {
...
       config.dirs.translations ? await this._extractTranslations(locales) : {};
 
+    console.log('Registering Handlebars helpers');
+    this._registerHelpers();
+
     const pageSets = GENERATED_DATA.getPageSets();
     for (const pageSet of pageSets) {
+      // Pre-process partials and register them with the Handlebars instance
+      const locale = pageSet.getLocale();
+      const translator = await Translator
```
tmeyer2115, line 110: Are we potentially creating multiple Translators for the same locale here? If so, maybe we should move this out of the pageSet loop as we only need one Translator per locale.

src/commands/build/sitesgenerator.js
```diff
@@ -97,12 +95,35 @@ exports.SitesGenerator = class {
...
+        .create(locale, GENERATED_DATA.getLocaleFallbacks(locale), translations);
+      const partialPreprocessor = new PartialPreprocessor(translator)
+
+      console.log(`Registering Handlebars partials for locale ${locale}`);
+      for (const partial of partialRegistry.getPartials()) {
+        hbs.registerPartial(partial.getName(), partialPreprocessor.process(partial.getFileContents()));
```
tmeyer2115, line 116: LL

src/commands/build/sitesgenerator.js
```diff
@@ -97,12 +95,35 @@ exports.SitesGenerator = class {
...
+      }
+
+      // Pre-process page template contents - these are not registered with the Handlebars instance,
+      // the PageWriter compiles them with their args
+      for (const page of pageSet.getPages()) {
+        const processedPartial = partialPreprocessor.process(page.getTemplateContents());
```
tmeyer2115, line 122: In HBS terminology, these pages are 'templates'. Is a 'template' considered a 'partial'? If not, maybe we rename the preprocessor. I guess someone could use these "helpers" in their page templates. But, they may bork their stuff if the "helpers" resolve to an SDK call.

src/models/partialsregistry.js
```diff
@@ -0,0 +1,81 @@
...
+   * the path's root is included in the partial naming scheme; the relativePaths
+   * are considered relative to the given path rather than root.
+   *
+   * @param {Array<String>} fullyQualifiedPaths The set of partials to register
+   *                                      using full path for the naming scheme.
+   * @param {Array<String>} relativePaths The set of partials to register using
+   *                                      the relative path for the naming scheme.
+   * @returns {PartialsRegistry}
+   */
+  static build({ fullyQualifiedPaths, relativePaths }) {
```
tmeyer2115, line 33: Do you think it would make more sense for this method to take in the Theme directory and an array of custom partials? My thinking is that would make this interface a bit more friendly conceptually. We get partials from the Theme and any custom partials. The user of the method doesn't need to know the implementation detail that partials from the Theme don't have a fully qualified name.

alexisgrow: Yeah totally! Updating!

src/commands/build/sitesgenerator.js
alexisgrow: Right now we have a domain model called a "PageTemplate" which is a domain model of a partial for a page. I'm thinking it may be clearer if we renamed "PageTemplate" to "PagePartial", would that be better?

src/commands/build/sitesgenerator.js
alexisgrow: We can move it! Currently there is only one pageSet per locale, but that may not always be the case in the future.

src/commands/build/sitesgenerator.js
cea2aj: My understanding is that in HBS terminology, all partials are templates, but not all templates are partials, so I like the idea of renaming to "PagePartial"

alexisgrow: Awesome, I opened a new PR to do the renaming since it's a ton of lines changed. You can find it here: https://github.com/yext/jambo/pull/86

alexisgrow: Renamed, now this line is `const processedPartial = partialPreprocessor.process(page.getPartialContents());`

tmeyer2115: Sorry for the confusion, I was advocating that we change the name of the PartialPreprocessor class itself. Because it operates on not just partials, but templates too, I think it should have a more inclusive name. I'm thinking HandlebarsPreprocessor. We can do that as another item.

PR #85, opened 2020-08-18
translator.js: pass locale to runtime helpers to handle multiple plurals

oshi97:
This commit changes translator.js to pass through locale to the runtime
translation helpers, so that the proper plural form can be chosen from
a dynamically generated count (since this depends on both the count
and the locale). The SDK will be getting the locale specific logic
for choosing the correct plural form. Since the first key in i18next
and also gettext is indexed as 0, translatePlural has also been
updated to set the first translation key to 0 instead of 1, and
the a key that ends with '_plural' is now has key '1' instead of 'plural' to
be more consistent with gettext plural forms.

TEST=auto

Add unit tests for a locale with multiple plurals (lt-LT)


tests/i18n/translator/translator.js
```diff
@@ -3,95 +3,141 @@ const path = require('path');
...
-  }
-
-  translator = await Translator.create('fr-FR', ['fr'], translations);
-});
-
-describe('Translations without pluralization or context', () => {
-  it('simple translation works as expected', () => {
-    const translation = translator.translate('Item');
-    expect(translation).toEqual('Article');
+describe('translations with one plural form (French)', () => {
```
oshi97, line 6: I did not add any new tests to this french describe block,  it's just indented in one extra level

src/i18n/translator/translator.js
```diff
@@ -35,11 +35,12 @@ class Translator {
    * 
    * @param {string} phrase The phrase to translate.
    * @param {string} pluralForm The untranslated, plural form of the phrase.
+   * @param {string} originalLocale The original locale of the passed in phrase.
```
alexisgrow, line 38: ooh I saw there's an item in for this week that mentioned we'll have localized versions of the SDK, are we having a multi-lang version as well?

oshi97: I don't think there's any plan for a multi-language version, but I was thinking it would be nicer to tell the SDK that the fallback language is english (maybe it shouldn't be a parameter), instead of having the SDK have logic saying that if there's no locale specified then assume it's english.

tests/i18n/translator/translator.js
```diff
@@ -3,95 +3,141 @@ const path = require('path');
...
+      '1 location selected',
+      '[[count]] locations selected]'
+    );
     const expectedResult = {
-      1: 'Missing [[count]] translation [[name]]',
-      plural: 'Missing [[count]] translations [[name]]'
+      0: 'Pasirinkta [[count]] tinklalapis',
+      1: 'Pasirinkta [[count]] tinklalapiai',
+      2: 'Pasirinkta [[count]] tinklalapių',
+      locale: 'lt-LT'
```
cea2aj, line 138: So the pluralizations will be keyed differently based on locale?

oshi97: yeah, locales with multiple plural forms need more keys

PR #86, opened 2020-08-19
Rename domain model PageTemplate to PagePartial

alexisgrow:
Updating to address a comment on this PR: https://github.com/yext/jambo/pull/83

TEST=unit


src/commands/build/pagepartialdirector.js
```diff
@@ -0,0 +1,121 @@
+const PagePartial = require('../../models/pagepartial');
```
alexisgrow, line 1: This is all name changes, so the diff is MUCH easier to review if you go commit by commit; I organized this so the first commit renames all the usages and the second commit renames the files!

src/commands/build/pagepartialdirector.js
```diff
@@ -33,18 +33,18 @@ module.exports = class PagePartialDirector {
...
   direct(pagePartials) {
-    const pageNameToTemplates = this._getPageNameToTemplates(pagePartials);
+    const pageNameToPartials = this._getPageNameToPartials(pagePartials);
 
     let localizedPagePartials = {};
     for (const locale of this._locales) {
       localizedPagePartials[locale] = [];
-      for (const templates of Object.values(pageNameToTemplates)) {
+      for (const templates of Object.values(pageNameToPartials)) {
         const pagePartial = this._findPagePartialForLocale(locale, templates);
```
cea2aj: Are these usages of "templates" intentional here?

src/commands/build/pagepartialdirector.js
```diff
@@ -82,20 +82,20 @@ module.exports = class PagePartialDirector {
...
    */
-  _getPageNameToTemplates(templates) {
+  _getPageNameToPartials(templates) {
     if (!templates || templates.length < 1) {
       return {};
     }
 
-    let pageNameToTemplates = {};
+    let pageNameToPartials = {};
     for (const pagePartial of templates) {
```
cea2aj: Here are some more usages of "templates", but I think it works fine here

src/commands/build/pagepartialdirector.js
alexisgrow: Updated!

src/commands/build/pagewriter.js
```diff
@@ -46,7 +46,7 @@ module.exports = class PageWriter {
         pageNameToConfig: pageSet.getPageNameToConfig(),
       });
 
-      const template = hbs.compile(page.getTemplateContents());
+      const template = hbs.compile(page.getPartialContents());
       const outputHTML = template(templateArguments);
```
cea2aj, line 50: These are the last usages of "template", but I think its fine inside this class

alexisgrow: Yeah I agree, I left these ones intentionally since in this is where we're compiling it as a Handlebars template!

PR #87, opened 2020-08-19
Add error message if missing all page templates for a locale

alexisgrow:
Previously this was throwing a "Cannot read property 'find' of undefined" error. Updating to print a helpful message and continue without generating a page set when missing page templates.

TEST=manual

Run `jambo build` locally for a site without page templates for a given locale.

### DIFF
--- a/src/commands/build/pagesetsbuilder.js
+++ b/src/commands/build/pagesetsbuilder.js
@@ -59,12 +59,18 @@ module.exports = class PageSetsBuilder {
 
     const pageSets = [];
     for (const [locale, pageConfigs] of Object.entries(localeToPageConfigs)) {
+      const partials = localeToPagePartials[locale];
+      if (!partials) {
+        console.log(`Warning: No page partials found for given locale '${locale}', not generating a page set for '${locale}'`);
+        continue;
+      }
+
       const localizedGlobalConfig = new GlobalConfigLocalizer(this._localizationConfig)
         .localize(this._globalConfig, locale);
 
       pageSets.push(new PageSet({
         locale: locale,
-        pages: this._buildPages(pageConfigs, localeToPagePartials[locale]),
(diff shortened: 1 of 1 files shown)


src/commands/build/pagesetsbuilder.js
```diff
@@ -59,12 +59,18 @@ module.exports = class PageSetsBuilder {
 
     const pageSets = [];
     for (const [locale, pageConfigs] of Object.entries(localeToPageConfigs)) {
+      const partials = localeToPagePartials[locale];
+      if (!partials) {
```
tmeyer2115, line 63: How would someone encounter this error? Does this check happen after we do the page template fallbacks/defaulting?

alexisgrow: Yep! This would happen when there are files in the `config/` directory for a page, but no corresponding files in the `pages/` directory (fallbacks included).

tmeyer2115: Gotcha!

tmeyer2115: Does the `localeToPagePartials[locale]` object contain the locale's templates for every page? If so, it seems like we're checking here that the `pages/` directory is effectively empty for the locale. Do we have another check that individual page templates are missing? For instance, what if there's no page template for the index page in any locale (but we do have templates for other pages)?

alexisgrow: >  it seems like we're checking here that the pages/ directory is effectively empty for the locale

Yes that's exactly right! 

> Do we have another check that individual page templates are missing?

Yes! That check happens when determining fallbacks, so at this point, we've got an Object "localeToPagePartials" mapping for each locale managed by the locale_config to its page templates. So for example, this could look like:

en: index
es: index, locations
fr: index
fr_CH: index, locations, products, faqs

PR #92, opened 2020-08-20
Update nomenclature for Handlebars partials and templates

alexisgrow:
Handlebars templates are not registered with Handlebars, they are the entry point for the HBS compilation. By that definition, Jambo, the pages/page.html.hbs files are templates. On the other hand, partials are registered with Handlebars and must have a name. We're currently using the term "PagePartials" when a more correct term is "PageTemplates", similarly, we're using "PartialPreprocessor" to process Handlebars content for both templates and partials, so it could be called "HandlebarsPreprocessor". (See this PR comment: https://github.com/yext/jambo/pull/90#pullrequestreview-471650562, and this comment: https://github.com/yext/jambo/pull/83#discussion_r473263671) Updating usages, variable names, log messages, and comments.

TEST=unit,manual

Run jambo build on i18n'd site, confirm it looks the same before and after this change.


src/commands/build/pagesetsbuilder.js
```diff
@@ -4,12 +4,12 @@ const Page = require('../../models/page');
 const PageConfig = require('../../models/pageconfig');
 const PageConfigDecorator = require('./pageconfigdecorator');
 const PageSet = require('../../models/pageset');
-const PagePartial = require('../../models/pagepartial');
-const PagePartialDirector = require('./pagepartialdirector');
+const PageTemplate = require('../../models/pagetemplate');
```
alexisgrow, line 7: Looking at the diffs commit-by-commit helps a *ton* for this PR since we're renaming files and changing a bunch of usages. I separated out the file renames from the content changes so the commit-by-commit diffs are pretty clean!

src/commands/build/pagewriter.js
```diff
@@ -26,6 +26,7 @@ module.exports = class PageWriter {
    * Writes a file to the output directory per page in the given PageSet.
    *
    * @param {PageSet} pageSet the collection of pages to generate
+   * @param {Object<String,String>} pageNameToTemplateContents
```
tmeyer2115, line 29: Does this belong here? Looks like the method only takes on param

src/commands/build/pagewriter.js
```diff
@@ -34,11 +35,7 @@ module.exports = class PageWriter {
     console.log(`Writing files for '${pageSet.getLocale()}' locale`);
 
     for (const page of pageSet.getPages()) {
-      if (!page.getConfig()) {
```
tmeyer2115, line 37: Was this error check moved elsewhere?

src/templates/templatepreprocessor.js
```diff
@@ -1,36 +1,36 @@
...
 
 /**
- * This class performs preprocessing on partials before they are registered
- * with Handlebars. The preprocessing is applicable to any type of partial
+ * This class performs preprocessing on templates before they are registered
+ * with Handlebars. The preprocessing is applicable to any type of template
  * used with Jambo.
  */
-class PartialPreprocessor {
+class TemplatePreprocessor {
```
tmeyer2115, line 8: This should run on both templates and partials. We want a name that encompasses both. Maybe HandlebarsPreprocessor? (Since both are HBS)

alexisgrow: Updated!

src/commands/build/pagewriter.js
alexisgrow: This check is superfluous since this error is caught significantly before this point, but I think removing it probably doesn't belong in this renaming PR! I'm going to remove the change from this PR and we can address in a separate one!

PR #93, opened 2020-08-20
Update generation to support backwards compatiblity with locale naming

alexisgrow:
There was an issue where the localized-pages branch was not backwards compatible if the locale-specific naming scheme was used in page files for locales that *didn't* match the locale in the global_config. For example, if the locale in my global_config was "en", but I had a page files: `config/name.es.json` and `page/page.es.html.hbs`, they would be considered files for the "es" locale, and would not be generated, even though we didn't have a locale_config. Eliminated defaulting to the global_config's locale since we were only using it for a key in an Object and it was complicating the fallback logic.

TEST=manual,unit

Tested on a repo with files using the locale-specific naming scheme without a locale_config.json file; verified that the following pages were generated: locations.es.html, index.es.html, index.html, with the following warnings: 
- Warning: No page template 'index.fr', not generating a 'index.fr' page
- Warning: No page template 'locations.fr', not generating a 'locations.fr' page

Tested on the same repo with a locale_config.json file; verified that the following pages were generated: es/index.html, es/locations.html, fr/override/index.html, index.html, with the following warning:
- Warning: No page template 'locations' found for 'fr' locale, not generating a 'locations' page found for 'fr' locale

### DIFF
--- a/src/commands/build/pagesetsbuilder.js
+++ b/src/commands/build/pagesetsbuilder.js
@@ -13,12 +13,7 @@ const PageTemplateDirector = require('./pagetemplatedirector');
  * of {@link PageSet}s.
  */
 module.exports = class PageSetsBuilder {
-  constructor({ defaultLocale, globalConfig, localizationConfig }) {
-    /**
-     * @type {String}
-     */
-    this._defaultLocale = defaultLocale;
-
+  constructor({ globalConfig, localizationConfig }) {
     /**
      * @type {GlobalConfig}
      */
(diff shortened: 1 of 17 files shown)


src/commands/build/pagesetsbuilder.js
```diff
@@ -39,29 +35,20 @@ module.exports = class PageSetsBuilder {
...
     const pageSets = [];
     for (const [locale, pageConfigs] of Object.entries(localeToPageConfigs)) {
       const templates = localeToPageTemplates[locale];
-      if (!templates) {
-        console.log(`Warning: No page templates found for given locale '${locale}', not generating a page set for '${locale}'`);
+      if (!templates || templates.length < 1) {
+        const localeMessage = config.getLocale() !== NO_LOCALE
+          ? ` for '${locale}' locale`
+          : '';
+        console.log(`Warning: No page templates found${localeMessage}, not generating a page set${localeMessage}`);
```
cea2aj, line 51: nit: for warnings there is a console.warn() which prints to stderr. It's not necessary, but it could be useful to spruce if they want to be able to see just the errors.

alexisgrow: Ah yes I was thinking the same thing! For now, I'll merge as logs because the answers-hitchhiker-theme's `grunt watch` task currently fails if anything is printed to stderr (here: https://github.com/yext/answers-hitchhiker-theme/blob/master/static/Gruntfile.js#L38), so before making the change to log to stderr, we'd want to adjust that Gruntfile!

PR #96, opened 2020-08-21
Update translation extractor for new preprocessing helpers

oshi97:
This commit updates the translation extraction command (renamed from extract-i18n
to extract-translations) to use the new preprocessing translation helpers.
To do this, the TranslationExtractor class was reimplemented to use the
Handlebars built-in AST parser, instead of xgettext-template, which was not
able to match our use case.

J=SLAP-580
TEST=manual, auto

Updated unit tests
In a localized jambo repo with translation helpers, run jambo extract-translations
see that I can specify the output to a specific file (and that parent folders
will be created as needed) otherwise it will just default to messages.pot.
See that the pot file's contents as well as filepath:line number references are correct.

Currently you can only run jambo commands in the same folder as the jambo.json
config. If we update jambo to let you run it in any folder within the jambo repo,
this command will likely have to be updated to get the correct filepath:line number references
in the pot files.


src/commands/extract-translations/jambotranslationextractor.js
```diff
@@ -1,32 +1,29 @@
...
-      files: this.files,
-      directories: this.directories,
-      output: outputFile,
-      ignore: this.gitignorePaths
-    };
-    await extractTranslations(options);
-    console.log(`Extracted translations to ${outputFile}`);
+  async extract(outputPath) {
+    const { files, directories } = this._getFilesAndDirsFromJamboConfig();
+    const gitignorePaths = this._parseGitignorePaths();
```
cea2aj, line 19: nit: I think down the line _parseGitignorePaths and _getFilesAndDirsFromJamboConfig could be placed in different files than this one, but that's probably out of scope of this PR

src/i18n/extractor/translationextractor.js
```diff
@@ -0,0 +1,101 @@
...
+  getPotString() {
+    return this._extractor.getPotString();
+  }
+
+  /**
+   * Saves the currently extracted messages to a pot file with the designed path.
+   * Creates any parent directories as necessary.
+   * @param {string} outputPath
+   */
+  savePotFile (outputPath = 'messages.pot') {
```
cea2aj, line 55: nit: This default is defined here and in cli.js. Which one is the actual default?

src/commands/extract-translations/jambotranslationextractor.js
oshi97: oh yeah these methods should probably be refactored out in the future

src/i18n/extractor/translationextractor.js
oshi97: I wanted to write this in a portable way, so the JamboTranslationExtractor would just be a wrapper over the inner TranslationExtractor and apply some settings to it depending on the state of the jambo site its operating on. I was thinking it would be nice for the inner TranslationExtractor to have a default output file, but maybe it should just error out instead? I'll take out this inner default.

src/commands/extract-translations/jambotranslationextractor.js
cea2aj: Ya, I really like the idea of having a model for the jambo config

PR #97, opened 2020-08-21
Require default locale in locale_config

alexisgrow:


src/models/localizationconfig.js
```diff
@@ -27,6 +28,13 @@ module.exports = class LocalizationConfig {
      */
     this._localeToConfig = config.localeConfig || {};
 
+    const shouldHaveConfigForDefault =
+      this._defaultLocale !== NO_LOCALE && Object.keys(this._localeToConfig).length > 0;
```
tmeyer2115, line 32: Shouldn't there be config in `localeToConfig` for the default as long as it's not a legacy site (using only global_config)? With this boolean, if `localeToConfig` was just an empty object, but I had a default locale, would `shouldHaveConfigForDefault` be false?

alexisgrow: Ah, so is having a "default" locale in the locale_config required? Currently, this code verifies that if there is a "default" locale specified, it has config. Should it be changed to require that a default exists?

tmeyer2115: If the `defaultLocale` was present, it wouldn't be `NO_LOCALE`, right? In that case, if `localeToConfig` was an empty object, `shouldHaveConfigForDefault` would evaluate to false, even though `defaultLocale !== NO_LOCALE`. If I had a locale_config file, could I trigger this behavior by having a `defaultLocale` and no `localeToConfig` object? I think in that case, we'd want an error telling them to add config for the default locale. Yeah, a `defaultLocale` is required in the config!

alexisgrow: Sweet, if a `defaultLocale` is required in the config then this check should definitely be different, I'll update!

PR #98, opened 2020-08-21
TranslateInvocation: update to use the Handlebars parser

oshi97:
This commit updates the TranslateInvocation model to use the
Handlebars parser for parsing out properties of the invocation.

J=SLAP-599
TEST=auto, manual

wrote unit tests for TranslateInvocation

added a page with contents `{{{read 'partials/temp'}}}`
and partials/temp.hbs with contents 

```hbs
{{translate phrase='Hello my name is [[name]]' name=card.name}}
blahasdf
```
on running jambo build got result 

```hbs
{{ runtimeTranslation phrase='Bonjour, mon nom est [[name]]' name=card.name }}
blahasdf
```

### DIFF
--- a/src/handlebars/models/translateinvocation.js
+++ b/src/handlebars/models/translateinvocation.js
@@ -95,25 +95,57 @@ class TranslateInvocation {
    * @returns {TranslateInvocation} The resulting {@link TranslateInvocation}.
    */
   static from(invocationString) {
-    const tree = Handlebars.parse(invocationString);
-    if (!tree.body || tree.body.length !== 1) {
+    let tree;
+    try {
+      tree = Handlebars.parse(invocationString);
+    } catch (err) {
       throw new UserError(
-        `Error: Could not parse ${invocationString} as a valid translate helper.`, err.stack);
+        `Error: Could not parse "${invocationString}" as a valid translate helper.`, err.stack);
     }
-    const invokedHelper = tree.body[0].path.original;
-    const hashPairs = tree.body[0].hash.pairs;
-    const parsedParams = this._convertHashPairsToParamsMap(hashPairs);
+    if (tree.body.length !== 1) {
+      throw new UserError(
+        `Error: "${invocationString}" has multiple handlebars nodes.`);
+    }
+    const node = tree.body[0];
+    if (node.type !== 'MustacheStatement') {
+      throw new UserError(
+        `Error: "${invocationString}" must be a MustacheStatement.`);
+    }
+    return this._fromMustacheStatementNode(node, invocationString);
+  }
+
+  /**
+   * Creates a {@link TranslateInvocation} from a Handlebars MustacheStatement.
+   * @param {MustacheStatement} mustacheStatement
+   * @param {string} invocationString
+   */
+  static _fromMustacheStatementNode(mustacheStatement, invocationString) {
+    const invokedHelper = mustacheStatement.path.original;
+    const hashPairs = mustacheStatement.hash.pairs;
+    const parsedParams = this._convertHashPairsToParamsMap(hashPairs, invocationString);
     return new TranslateInvocation(invokedHelper, parsedParams);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/handlebars/models/translateinvocation.js
```diff
@@ -86,30 +88,35 @@ class TranslateInvocation {
...
-        // Strip the wrapper '' for string params
-        const paramValue = paramOperands[1] && paramOperands[1].replace(/(^')|('$)/g, '');
-        params[paramName] = paramValue;
-        return params;
-      }, {});
-
+    const tree = Handlebars.parse(invocationString);
+    if (!tree.body || tree.body.length !== 1) {
+      throw new UserError(
+        `Error: Could not parse ${invocationString} as a valid translate helper.`, err.stack);
```
tmeyer2115, line 101: Is `err` defined here?

src/handlebars/models/translateinvocation.js
```diff
@@ -86,30 +88,35 @@ class TranslateInvocation {
...
+    }
+    const invokedHelper = tree.body[0].path.original;
+    const hashPairs = tree.body[0].hash.pairs;
+    const parsedParams = this._convertHashPairsToParamsMap(hashPairs);
```
tmeyer2115, line 105: Do we want to wrap any of this in a try/catch? Is there any way this could blow up (body doesn't have path, body doesn't have hash)?

oshi97: Added some validation based on the types at https://github.com/handlebars-lang/handlebars.js/blob/master/docs/compiler-api.md
added unit test coverage for everything I could think of

src/handlebars/models/translateinvocation.js
```diff
@@ -86,30 +88,69 @@ class TranslateInvocation {
...
+      tree = Handlebars.parse(invocationString);
+    } catch (err) {
+      throw new UserError(
+        `Error: Could not parse "${invocationString}" as a valid translate helper.`, err.stack);
+    }
+    if (tree.body.length === 0) {
+      throw new UserError(`Error: "${invocationString}" does not have any handlebars nodes.`);
+    }
+    if (tree.body.length > 1) {
+      throw new UserError(`Error: "${invocationString}" has multiple handlebars nodes.`);
```
tmeyer2115, line 109: I think the errors in this method and the ones below will be difficult for people to parse. Understanding them requires some detailed Handlebars knowledge (what a node is, what a subexpression is, etc.). For a consumer of this system, they may only understand Handlebars at a high level. That is, they'll understand "${invocationString} is not a properly formatted Handlebars helper call", but not much else. What would you think of having the `from` method be like you had before, but wrapped in a big ol try catch. We'd throw a single, high-level error, like the one I mentioned. Do you think that will be enough to help people debug?

oshi97: Ah, yeah there's probably not that much point to go into this detail and it also just makes the code a lot simpler

src/handlebars/models/translateinvocation.js
```diff
@@ -86,30 +88,59 @@ class TranslateInvocation {
...
+   */
+  static _convertHashPairsToParamsMap (hashPairs) {
+    return hashPairs.reduce((map, pair) => {
+      const expression = pair.value;
+      if (expression.type === 'NullLiteral') {
+        map[pair.key] = 'null';
+      } else if (expression.type === 'UndefinedLiteral') {
+        map[pair.key] = 'undefined';
+      } else if (expression.type === 'SubExpression') {
+        throw new UserError(
```
tmeyer2115, line 136: Should we just do a normal throw new Error here as well?

oshi97: I feel like this should be a UserError since this is just to prevent people from doing stuff like
```hbs
{{ translate phrase='[[abc]]' abc=(concat 'a' 'b' 'c') }}
```
(I was thinking it could possibly be something we support in the future? the very far future)

PR #99, opened 2020-08-24
dont wrap static {{translate}} results with quotes

oshi97:
When there is a static translation in hbs, we dont want to wrap it in quotes.
Originally, this item was going to also not wrap static translateJS results
in quotes as well, but after taking another look that seemed confusing, since
translateJS would return a javascript string if it was a runtime translation,
but the raw(?) text if it was a static translation. In order to know whether
to wrap a translate helper in quotes you would have to think about whether
this was something that would be statically translated i.e. (no interpolation or
plurals). This seemed like something that could end up very frustrating.

Instead, with this pr {{translateJS}} will always compile to a js string, and
{{translate}} will always compile down to raw text without quotes,
whether runtime or not.

J=SLAP-592
TEST=auto

updated unit tests

### DIFF
--- a/src/handlebars/handlebarspreprocessor.js
+++ b/src/handlebars/handlebarspreprocessor.js
@@ -55,10 +55,9 @@ class HandlebarsPreprocessor {
           invocation.getPhrase(), translationContext) :
         this._translator.translate(invocation.getPhrase());
     }
-    translatorResult = `'${translatorResult}'`;
 
     if (invocation.canBeTranslatedStatically()) {
-      return translatorResult;
+      return invocation.getInvokedHelper() === 'translateJS' ? `'${translatorResult}'` : translatorResult;
     }
     const interpParams = invocation.getInterpolationParams();
 
@@ -74,6 +73,8 @@ class HandlebarsPreprocessor {
    * Constructs a call to the SDK's Javascript method for run-time translation.
    * This call is constructed using the translation(s) for a phrase and any interpolation
    * paramters.
+   * TODO: after updating translate invocation to use the Handlebars parser, update this method
+   * to escape single quotes in translatorResult.
    *
    * @param {Object|string} translatorResult The translation(s) for the phrase.
    * @param {boolean} needsPluralization If pluralization is required when translating.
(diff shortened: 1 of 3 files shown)


src/handlebars/handlebarspreprocessor.js
```diff
@@ -55,10 +55,9 @@ class HandlebarsPreprocessor {
           invocation.getPhrase(), translationContext) :
         this._translator.translate(invocation.getPhrase());
     }
-    translatorResult = `'${translatorResult}'`;
 
     if (invocation.canBeTranslatedStatically()) {
-      return translatorResult;
+      return invocation.getInvokedHelper() === 'translateJS' ? `'${translatorResult}'` : translatorResult;
```
alexisgrow, line 60: Do you think Jambo should add quotes to the translateJS helper? I can see it being hard to remember for an end user that they would need quotes wrapping `'{{ translate phrase='Hello' }}'` but not `{{ translateJS phrase='Hello' }}`. On the other hand, it also may be convenient for them not have quotes if they only use translateJS?

oshi97: It just felt a little weird that translateJS might or might not need quotes depending on whether it would be statically 
 translatable. Since for something like
```js 
const hello = '{{ translateJS phrase='Hello' }}'; 
```
```js
const nonStaticHello = {{ translateJS phrase='Hello' pluralForm='Hellooooo'}}; 
```

A hitchhiker would have to remember to not wrap the second one in quotes, since it compiles down to ANSWERS.translateJS('...stuff that goes in here...'), since if that it was wrapped in single quotes it would cause an error, 
```js
const nonStaticHello = 'ANSWERS.translateJS('...stuff that goes in here...')';
```
and it was wrapped in double quotes it would turn into the string (I'm really overusing this formatting lol)

```js
const nonStaticHello = "ANSWERS.translateJS('...stuff that goes in here...)";
```
Idk it seemed easier to remember that translateJS will give you a javascript string and translate will straight up change your source code, as opposed to translateJS will give you a javascript string if it is a runtime translation and raw source code if it is static, and translate will always give you raw source code. I couldn't really think of a reason why you would want to translate your raw source code in js either.

alexisgrow: Ahhh, yes, since it's transpiling to Answers SDK JS that totally makes sense! It does seem easier to say "translateJS returns a string" than "translateJS needs to be wrapped in quotes when the call can be statically translated, otherwise wrapping it in quotes breaks it".

PR #100, opened 2020-08-24
Intermixed markdown

cea2aj:
Add tests that confirm that the translator handles intermixed markdown as expected

J=SLAP-581
TEST=auto, manual

In addition to automatic tests, I also tested the handlebars translate helper in an internationalized jambo site. The mixed markdown fails due to the regex not picking characters including "<" and ">". Oliver's switch to the handlebars parser fixes this issue: https://github.com/yext/jambo/pull/98

### DIFF
--- a/src/i18n/translator/translator.js
+++ b/src/i18n/translator/translator.js
@@ -148,7 +148,7 @@ class Translator {
    * @param {Array<string>} fallbacks A prioritized list of translation fallbacks
    *                                  for the locale.
    * @param {string} translations The translations for the locale and associated fallbacks.
-   * @param {Object<string, Object} A map of locale to translations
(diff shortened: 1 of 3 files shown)


src/i18n/translator/translator.js
```diff
@@ -148,6 +148,7 @@ class Translator {
    * @param {Array<string>} fallbacks A prioritized list of translation fallbacks
    *                                  for the locale.
    * @param {string} translations The translations for the locale and associated fallbacks.
+   * @param {Object<string, Object>} A map of locales to translations
```
oshi97, line 151: don't forget to remove the defunct jsdoc line above this one!!! Also the first word of the description should be the variable name (translations)

cea2aj: My bad, fixed!

PR #101, opened 2020-08-24
Throw errors on page generation collisions

alexisgrow:
TEST=manual,unit
J=SLAP-598

Test triggering the "duplicate pageName + locale" error by having two page configs for the same page: index.es.json and index.json (default locale 'es'). Test triggering the output path collision error by using the same urlFormat for all locales in my locale_config.json file. Test that without the intentional errors, jambo build succeeds.


src/commands/build/pageuniquenessvalidator.js
```diff
@@ -0,0 +1,83 @@
...
+        .map(page => `[${page.getName()}, ${page.getLocale()}]`)
+        .join(' and ');
+
+      brokenRuleDescription.push(`\tPages ${pageNameLocaleCombinations} are configured to use output path '${path}'`);
+    };
+
+    return brokenRuleDescription.join('\n');
+  }
+
+  /**
```
oshi97, line 69: is there some general place this method can find its home? It seems useful and isn't really specific to this file.

src/commands/build/pageuniquenessvalidator.js
```diff
@@ -0,0 +1,83 @@
...
+   * Runs a set of validation tests on the given pages. Throws an error with a description of the
+   * broken validation rule if present.
+   *
+   * @param {Array<Page>} pages
+   */
+  validate(pages) {
+    if (!pages || pages.length < 1) {
+      return;
+    }
+    const brokenValidationRule = this._validatePageNameLocaleCombinations(pages)
```
oshi97, line 18: would it be simpler to just have the validationMethods like this._validatePageNameLocaleCombinations throw the errors themselves? That way in the future we don't have to keep glueing more validation calls to this brokenValidationRule variable and can just call them in series.

src/commands/build/pageuniquenessvalidator.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * Validates the [pageName, locale] combinations in the given pages. Returns a description of the
+   * broken validation rule if present.
+   *
+   * @param {Array<Page>} pages
+   * @returns {String}
+   */
+  _validatePageNameLocaleCombinations(pages) {
```
oshi97, line 33: Would it be possible to get some simple unit tests for these 2 methods?

src/commands/build/pageuniquenessvalidator.js
alexisgrow: Yes! This sprint we've got a follow-up item to add a top-level validator class (https://yexttest.atlassian.net/browse/SLAP-216), so my understanding is that this will be moved as part of that.

src/commands/build/pageuniquenessvalidator.js
alexisgrow: Yes totally agree, updating!

src/commands/build/pageuniquenessvalidator.js
alexisgrow: Totally! Added!

src/models/generateddata.js
```diff
@@ -70,6 +71,11 @@ module.exports = class GeneratedData {
       globalConfig: globalConfig,
     }).build(pageConfigs, pageTemplates);
 
+    const allPages = pageSets
+      .flatMap(pageSet => pageSet.getPages());
+
+    new PageUniquenessValidator().validate(allPages);
```
tmeyer2115, line 77: I'd suggest separating the creation of GeneratedData from its validation. By including the validation here, we're being prescriptive of how GeneratedData should be used. Given that we have only one consumer of GeneratedData now and one validation step, this isn't a huge deal. In the future, though, we might have multiple consumers of GenreatedData and they may have different ideas about what's an acceptable state for the GeneratedData. In some sense, GeneratedData is just a dumb data container, you can put in whatever you'd like. If your usage of GeneratedData requires additional restrictions, your consumer should be responsible for imposing those restrictions. 

I'd say just do this validation when we call the static from method. How does that sound?

alexisgrow: Sure, that SGTM, I'll update!

PR #104, opened 2020-08-24
Create translator for each PageSet even if no locale config if provided

alexisgrow:
Since single-locale sites don't have a translator they could return an error much later in the HandlebarsProcessor `Cannot read property 'translate' of undefined`. Even same-language content could be used with the translate helpers for interpolation or pluralization, so we can't simply skip the pre-processing. Instead, we will always require a translator for pre-processing, even if the translator has no translations.

TEST=manual

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -116,12 +116,6 @@ exports.SitesGenerator = class {
       ? await this._extractTranslations(locales, configRegistry.getLocalizationConfig())
       : {};
 
-    const localeToTranslator = {};
-    for (const locale of locales) {
-      localeToTranslator[locale] = await Translator
-        .create(locale, GENERATED_DATA.getLocaleFallbacks(locale), translations);
-    }
-
     // Register needed Handlebars helpers.
     console.log('Registering Jambo Handlebars helpers');
     try {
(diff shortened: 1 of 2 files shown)


alexisgrow:
Are there cases where there wouldn't be a translator? For example, an US based developer working on an English site?

alexisgrow:
> I think, even in the case of one locale, we should have a translator with the preprocessor. As an example, I might be a US based developer working on a French site. I have English strings in my various translate helpers, but they do need to be translated to French for the actual site.

Talked offline! There are same language cases like English -> English (or any other language), where we want to preprocess Handlebars content without translating anything. Because of that, we don't want to short-circuit this method and return the content passed in. Instead, we will require a translator be provided, even if the translator doesn't translate any content.

src/commands/build/sitesgenerator.js
```diff
@@ -134,7 +128,9 @@ exports.SitesGenerator = class {
     for (const pageSet of pageSets) {
       // Pre-process partials and register them with the Handlebars instance
       const locale = pageSet.getLocale();
-      const handlebarsPreprocessor = new HandlebarsPreprocessor(localeToTranslator[locale]);
+      const translator = await Translator
```
oshi97, line 131: I think this creates the same translator once for each page in a locale, we can definitely optimize things further but do you think there's some way to avoid this in the short term as well?

oshi97: could also be that the i18next init stuff basically has no impact on build speed in which case the current code is fine as well! (actually I'm kind of curious how long Translator.create() takes now)

alexisgrow: Oh this creates the same translator for each page set - so there is a set of pages generated per-locale, which is what this is looping through!

oshi97: ohhh okay got it

PR #105, opened 2020-08-25
handlebarspreprocessor: make translate helper regex non-greedy

oshi97:
This commit makes the regex non-greedy so that it ends its match on the
first closing mustache found, instead of the last. This fixes both commented out
translation helpers and multiple translation helpers on a single line.

J=SLAP-591, SLAP-595
TEST=auto, manual

added unit tests
test jambo build locally with a commented out helper, and also two helpers on the same line

### DIFF
--- a/tests/handlebars/handlebarspreprocessor.js
+++ b/tests/handlebars/handlebarspreprocessor.js
@@ -40,7 +40,7 @@ describe('HandlebarsPreprocessor works correctly', () => {
     expect(handlebarsPreprocessor.process(rawHbsHandlebarsContent)).toEqual(processedHbsHandlebarsContent);
   });
 
-  it('passes correct arguments to translatePlural', () => {
+  describe('when translating simple plural english', () => {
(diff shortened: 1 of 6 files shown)


tests/handlebars/handlebarspreprocessor.js
```diff
@@ -40,7 +40,7 @@ describe('HandlebarsPreprocessor works correctly', () => {
     expect(handlebarsPreprocessor.process(rawHbsHandlebarsContent)).toEqual(processedHbsHandlebarsContent);
   });
 
-  it('passes correct arguments to translatePlural', () => {
+  describe('when translating simple plural english', () => {
```
alexisgrow, line 43: Would it make sense to say "when translating a language with a single form for plurals", i.e. is this specific to English?

oshi97: yeah that's better! I just wanted to group them so I didn't have top duplicate the mock implementation

PR #108, opened 2020-08-26
Build validation step for config files

cea2aj:
Add RawConfigValidator which runs validation on the raw config files. This lays a foundation that makes it easy to add validation to the config files for jambo builds. Current validation includes checking that global_config.json exists, that "default" and "localeConfig" keys are present if local_config.json is present, and a check that there exists a localeConfig for every locale referenced by the page config file names. I also moved some utils to configutils.js and added enums to constants.js

Added unit tests for the validators

J=SLAP-590
TEST=manual, auto

Ran unit tests. Also manually modified a jambo site to observe the validation correctly pass or fail. Successfully build an internationalized jambo site.


cea2aj:
What do you all think of this validation architecture? Should I pull the RawConfigValidator out of SitesGenerator.generate() and put it in it in a SitesGenerator validation function which must pass before generate() will execute?

cea2aj:
Tom brought up some old validation code which used a validation factory. Would that pattern be useful here?

src/commands/build/validation/globalconfigvalidator.js
```diff
@@ -0,0 +1,24 @@
...
+    this._globalConfig = globalConfig;
+  }
+
+  validate () {
+    this._validateConfigExists();
+  }
+
+  _validateConfigExists () {
+    if (!this._globalConfig) {
+      throw new UserError(`Error: Cannot find config/${fileNames.GLOBAL_CONFIG}, exiting.`);
```
alexisgrow, line 21: Since the `config/` directory is customizable in Jambo, would it be possible to omit it from this error message or use the config directory from the jambo.json file?

src/commands/build/validation/localeconfigvalidator.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+module.exports = class LocaleConfigValidator {
+  constructor (localizationConfig) {
+    /**
+     * @type {Object<string, string|Object>}
+     */
+    this._localizationConfig = localizationConfig;
+  }
+  
+  validate () {
```
alexisgrow, line 15: Do you think a JS doc on the `validate()` method in this class or the others would be helpful? For example, a JS doc here could say that this function performs a series of validation steps on the config and throws a `UserError` if any of them fail.

src/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,47 @@
...
+     * @type {Object<string, Object>}
+     */
+    this._configNameToRawConfig = configNameToRawConfig;
+  }
+
+  validate () {
+    new GlobalConfigValidator(this._getGlobalConfig()).validate();
+    if (!this._isMissingLocaleConfig()){
+      new LocaleConfigValidator(this._getLocaleConfig()).validate();
+      new PageConfigsValidator(this._configNameToRawConfig).validate();
```
alexisgrow, line 23: For this PageConfigsValidator call, would it make sense to filter out the non-page configs (`global_config` and `local_config`), like the GlobalConfigValidator and LocaleConfigValidator calls do?

src/commands/build/validation/globalconfigvalidator.js
cea2aj: Sure, I'll omit it

src/commands/build/validation/localeconfigvalidator.js
cea2aj: Ya, that works

src/commands/build/validation/rawconfigvalidator.js
cea2aj: Hmm. I passed in the configNameToRawConfig to the PageConfigsValidator constructor because the PageConfigsValidator checks that all the page configs have a configuration in locale_config.json, which means it needs access to both the page configs and a list of configured locales. What are your thoughts on filtering out the locale_config and global_config and passing in a list of configured locales as a second parameter in the constructor?

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,97 @@
...
+     * @type {Object<string, Object>}
+     */
+    this._pageConfigs = pageConfigs;
+
+    /**
+     * A list of the locales accociated with the page config files
+     * 
+     * @type {string[]}
+     */
+    this._pageLocales = this._getPageLocales();
```
alexisgrow, line 20: Would it make sense for this to live in `_validatePageLocalesHaveConfigs`?

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Get the locales associated with the page config files
+   * 
+   * @returns {string[]}
+   */
+  _getPageLocales () {
+    let locales = Object.keys(this._pageConfigs)
+      .filter(configName => 
+        configName !== configKeys.GLOBAL_CONFIG && 
+        configName !== configKeys.LOCALE_CONFIG &&
```
alexisgrow, line 48: Since only page configs are being passed in now, should these `LOCALE_CONFIG` and `GLOBAL_CONFIG` checks be removed?

tests/commands/build/validation/localeconfigvalidator.js
```diff
@@ -0,0 +1,44 @@
+const LocaleConfigValidator = require('../../../../src/commands/build/validation/localeconfigvalidator');
+const UserError = require('../../../../src/errors/usererror');
+
+describe('LocaleConfigValidator works properly', () => {
+  const config = {
```
alexisgrow, line 5: Since each of these consts are only used in one test below, could they be moved inside their test?

tests/commands/build/validation/localeconfigvalidator.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  const defaultMispelledConfig = {
+    defautl: 'en',
+    localeConfig: {}
+  }
+
+  const localeConfigMispelledConfig = {
+    defautl: 'en',
+    localeConfig: {}
```
alexisgrow, line 30: Locale config is present and spelled correctly here!

tests/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,61 @@
...
+const UserError = require('../../../../src/errors/usererror');
+
+describe('RawConfigValidator works properly', () => {
+  const multiLangConfig = {
+    global_config: {},
+    locale_config: {
+      default: 'en',
+      localeConfig: {
+        en: {
+          experienceKey: 'key',
```
alexisgrow, line 11: If parts of the sample data in this test and others aren't being validated/tested, like apiKey, expKey, urlFormat, can they be removed for now? (We can add it back later if we add validation for those things, but if they're superfluous it'd be great to remove.)

src/commands/build/validation/pageconfigsvalidator.js
cea2aj: Ya, that works

src/commands/build/validation/pageconfigsvalidator.js
cea2aj: Yep, I missed that

tests/commands/build/validation/localeconfigvalidator.js
cea2aj: fixing

tests/commands/build/validation/localeconfigvalidator.js
cea2aj: Sure

tests/commands/build/validation/rawconfigvalidator.js
cea2aj: Ya, that stuff isn't being validated currently. I'll remove it for now

src/commands/build/sitesgenerator.js
```diff
@@ -61,6 +62,9 @@ exports.SitesGenerator = class {
         }
       }
     });
+
+    new RawConfigValidator(configNameToRawConfig).validate();
```
oshi97, line 66: would it make sense to call this from inside ConfigurationRegistry? I think ideally all of this reading of config file stuff also goes in there (or like at least not directly in generate()) but that's a separate thing

src/constants.js
```diff
@@ -1 +1,11 @@
-exports.NO_LOCALE = '';
\ No newline at end of file
+exports.NO_LOCALE = '';
+
+exports.fileNames = {
+  GLOBAL_CONFIG: 'global_config.json',
```
oshi97, line 4: Think we probably want to replace usages of these hardcoded strings in other files as well. I think the convention for the exports is normally something like `FileNames` instead of `fileNames`

src/utils/configutils.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Parses the locale from a given configName
+ *
+ * @param {String} configName the file name of the config, without the extension
+ * @returns {String}
+ */
+parseLocale = function (configName) {
```
oshi97, line 7: I think we still want to specify these as const (or just do something like `exports.parseLocale = function (...` ) I didn't know you could create variables python style in js that's cool

tests/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,48 @@
...
+  it('does not throw an error with a correct single-language config', () => {
+    expect(() => new RawConfigValidator(singleLangConfig).validate()).not.toThrow(UserError);
+  });
+
+  it('_isMultiLanguageSite returns true for multilang sites', () => {
+    expect(new RawConfigValidator(multiLangConfig)._isMultiLanguageSite()).toEqual(true);
+  })
+
+  it('_isMultiLanguageSite returns false for single language sites', () => {
+    expect(new RawConfigValidator(singleLangConfig)._isMultiLanguageSite()).toEqual(false);
```
oshi97, line 46: nit but I think we prefer toBeTruthy and toBeFalsy calls here and above

tests/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const singleLangConfig = {
+    global_config: {},
+    index: {},
+    faqs: {},
+    people: {}
+  }
+
+  it('does not throw an error with a correct multi-language config', () => {
+    expect(() => new RawConfigValidator(multiLangConfig).validate()).not.toThrow(UserError);
```
oshi97, line 34: question but if an error that isn't of type UserError is thrown here does the test still fail?

src/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,75 @@
...
+   */
+  _getLocaleConfig () {
+    return this._configNameToRawConfig[configKeys.LOCALE_CONFIG];
+  }
+
+  /**
+   * @returns {Object<string, string|Object>} The keys are locale strings
+   */
+  _getPageConfigs () {
+    let pageConfigs = cloneDeep(this._configNameToRawConfig);
```
oshi97, line 59: (nit) const

src/commands/build/validation/globalconfigvalidator.js
```diff
@@ -0,0 +1,29 @@
+const UserError = require('../../../errors/usererror');
+const { fileNames } = require('../../../constants');
+
+/**
+ * Performs validation on global_config.json
+ */
+module.exports = class GlobalConfigValidator {
+  constructor (globalConfig) {
+    /**
```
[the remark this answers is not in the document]
cea2aj: thanks for the tip! fixing

src/commands/build/validation/globalconfigvalidator.js
```diff
@@ -0,0 +1,29 @@
...
+   * 
+   * @throws {UserError} Thrown if validation fails
+   */
+  validate () {
+    this._validateConfigExists();
+  }
+
+  _validateConfigExists () {
+    if (!this._globalConfig) {
+      throw new UserError(`Error: Cannot find ${fileNames.GLOBAL_CONFIG}, exiting.`);
```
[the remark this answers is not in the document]
cea2aj: Ya, I this code doesn't know where it's located so I'll updated the message as you suggested

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,92 @@
...
+   * Gets locales defined by page configs which do not
+   * have a localeConfig in locale_config.json
+   * 
+   * @param {string[]} pageLocales A list of locales defined by page config files
+   * @returns {string[]}
+   */
+  _getLocalesMissingConfigs (pageLocales) {
+    let locales = [];
+    pageLocales.forEach(locale => {
+      if (!(this._configuredLocales.includes(locale))){
```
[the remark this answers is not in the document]
cea2aj: done

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,92 @@
...
+    })
+    return locales;
+  }
+
+  /**
+   * @param {string[]} locales A list of locales which are missing configuration
+   */
+  _throwErrorForLocalesMissingConfigs (locales) {
+    if (locales.length == 1) {
+      throw new UserError(`The locale '${locales}' is defined in a page configuration file name but is not defined inside ${fileNames.LOCALE_CONFIG}`);
```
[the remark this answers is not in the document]
cea2aj: Updated

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Gets locales defined by page configs which do not
+   * have a localeConfig in locale_config.json
+   * 
+   * @param {string[]} pageLocales A list of locales defined by page config files
+   * @returns {string[]}
+   */
+  _getLocalesMissingConfigs (pageLocales) {
+    let locales = [];
+    pageLocales.forEach(locale => {
```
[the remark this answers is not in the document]
cea2aj: done

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,92 @@
...
+      )
+      .map(configName => {
+        return parseLocale(configName);
+      })
+
+    return this._removeDuplicates(locales);
+  }
+
+  /**
+   * @param {*[]} list
```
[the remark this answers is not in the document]
cea2aj: sure

src/utils/configutils.js
cea2aj: I was using the convention from utils/fileutils.js and utils/jamboconfigutils.js. Should I update those as well?

src/commands/build/validation/pageconfigsvalidator.js
```diff
@@ -0,0 +1,92 @@
...
+   * @returns {string[]}
+   */
+  _getPageLocales () {
+    let locales = Object.keys(this._pageConfigs)
+      .filter(configName => 
+        containsLocale(configName)
+      )
+      .map(configName => {
+        return parseLocale(configName);
+      })
```
alexisgrow, line 46: nit: Since these fit on the same line, can they be consolidated? e.g.
```js
      .filter(configName => containsLocale(configName))
      .map(configName => parseLocale(configName))
```

src/utils/configutils.js
oshi97: ohhh I see, lets just leave as is for now then we can codify our style rules later on

tests/commands/build/validation/localeconfigvalidator.js
```diff
@@ -0,0 +1,41 @@
+const LocaleConfigValidator = require('../../../../src/commands/build/validation/localeconfigvalidator');
+const UserError = require('../../../../src/errors/usererror');
+
+describe('LocaleConfigValidator works properly', () => {
+  it('does not throw an error with a correct config', () => {
+    const config = {
+      default: 'en',
+      localeConfig: {
+        en: {
+          experienceKey: 'key',
```
alexisgrow, line 10: Following up on this comment: https://github.com/yext/jambo/pull/108#discussion_r480092182, if parts of this test data aren't being tested, can they be removed?

tests/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,48 @@
...
+  });
```
alexisgrow, line 35: Should there be coverage for the validator testing an invalid config?

tests/commands/build/validation/rawconfigvalidator.js
```diff
@@ -0,0 +1,48 @@
...
+
+  it('does not throw an error with a correct single-language config', () => {
+    expect(() => new RawConfigValidator(singleLangConfig).validate()).not.toThrow(UserError);
+  });
+
+  it('_isMultiLanguageSite returns true for multilang sites', () => {
```
alexisgrow, line 41: Is this `_isMultiLanguageSite` function tested as a byproduct of the .validate() tests above?

src/commands/build/validation/rawconfigvalidator.js
cea2aj: fixed

tests/commands/build/validation/rawconfigvalidator.js
cea2aj: done

src/commands/build/sitesgenerator.js
cea2aj: ya, I like that idea

src/constants.js
cea2aj: Sure. Renaming to FileNames and ConfiKeys, and updating the hardcoded strings in the theme importer as well

tests/commands/build/validation/rawconfigvalidator.js
cea2aj: I don't think so. I'll switch it to Error just to be safe

src/commands/build/validation/pageconfigsvalidator.js
cea2aj: Sure

tests/commands/build/validation/localeconfigvalidator.js
cea2aj: Oops, missed this one. Fixing

tests/commands/build/validation/rawconfigvalidator.js
cea2aj: Ya, adding now

tests/commands/build/validation/rawconfigvalidator.js
cea2aj: Ya, I'll these multiLang tests

tests/validation/globalconfigvalidator.js
```diff
@@ -0,0 +1,13 @@
+const GlobalConfigValidator = require('../../src/validation/globalconfigvalidator');
+const UserError = require('../../src/errors/usererror');
+
+describe('GlobalConfigValidator works properly', () => {
+  it('does not throw an error with a correct config', () => {
+    const config = {};
+    expect(() => new GlobalConfigValidator(config).validate()).not.toThrow(UserError);
```
oshi97, line 7: I think we should just be checking that nothing is thrown at all (so just `.not.toThrow()` instead of `.not.toThrow(UserError)`, otherwise if there is some random programming error, lets say something isn't defined, these .not.toThrow tests will still pass. Technically I think you could just execute the validate method instead of wrapping it in an expect, but I kind of like how you wrapped it in an expect since it's slightly more clear what's being tested

cea2aj: Ya, that makes sense. For everything that should not throw, I'll make sure that nothing gets thrown at all. When something is supposed to throw, I'll check that the correct error gets thrown.

PR #109, opened 2020-08-27
Add translatePluralWithContext

cea2aj:
Add translatePluralWithContext to the translator class. Since this method shares code with translatePlural, I created some helper methods to reduce code duplication. I also added translateWithContext support in the preprocessor and escaped single quotes in the preprocessor translatorResult.

Tests added for the translator, preprocessor, and extractor.

J=SLAP-603
TEST=auto

Ran unit tests


src/i18n/translator/translator.js
```diff
@@ -29,6 +29,21 @@ class Translator {
...
+   * @param {string} phrase The phrase to translate.
+   * @param {string} context The context of the translation.
+   * @returns {string} The translated phrase or format string.
+   */
+  translateWithContext(phrase, context) {
+    const interpPlaceholders = this._getInterpolationPlaceholders(phrase);
+
+    return this._i18next.t(phrase, { context, ...interpPlaceholders});
+  }
+
```
cea2aj, line 46: I simply moved this method up

src/i18n/translator/translator.js
```diff
@@ -41,11 +56,9 @@ class Translator {
    *                                          see https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html
    */
   translatePlural(phrase, pluralForm, originalLocale = 'en') {
-    const escapedPhrase = phrase
-      .replace(/\[\[/g, '\\[\\[')
-      .replace(/\]\]/g, '\\]\\]');
+    const escapedPhrase = this._getEscapedPhrase(phrase);
     const pluralKeyRegex = new RegExp(`${escapedPhrase}_([0-9]+|plural)`);
-
+    const pluralizationIndex = 1; //Pluralization is at this index when phrase is split by the '_' delimiter
```
oshi97, line 61: I feel like this variable might be unnecessary. I think keys are the of the form (correct me if I'm wrong) key_context_plural or key_context or key_plural, so if the key is plural it should always be the last index?

src/i18n/translator/translator.js
```diff
@@ -80,18 +129,39 @@ class Translator {
...
-    return this._i18next.t(phrase, { context, ...interpPlaceholders});
+  /**
+   * Escapes the interpolation brackets in a phrase
+   * 
+   * @param {string} phrase 
+   * @returns {string}
+   */
+  _getEscapedPhrase(phrase) {
+    return phrase
+    .replace(/\[\[/g, '\\[\\[')
```
oshi97, line 163: nit but indents here

src/i18n/translator/translator.js
```diff
@@ -80,18 +129,39 @@ class Translator {
...
-   * interpolation, a translated format string, with the relevant placeholders,
-   * is returned.
+   * Creates a map of count (or 'plural') to the correct translated form
    * 
-   * @param {string} phrase The phrase to translate.
-   * @param {string} context The context of the translation.
-   * @returns {string} The translated phrase or format string.
+   * @param {string} locale The locale used when creating the map
+   * @param {RegExp} pluralRegex Regex that matches pluralized keys
+   * @param {string} singularPhrase The singular form of the phrase
```
oshi97, line 136: nit but I think `${phrase}_${context}` (when translating plural with context) is technically not the singular form of the phrase, maybe you could call this something like translationKey or maybe pass context and phrase as separate params?

tests/fixtures/handlebars/rawtemplate.hbs
```diff
@@ -8,4 +8,7 @@
         {{translateJS phrase='Mail now [[id1]]' context='Mail is a verb' id1=profile.name}}
     </script>
     <button>{{translate phrase='Mail now [[id1]]' context='Mail is a verb' id1=myName}}</button>
+    <div>
+        {{ translate phrase='The [[count]] person went on a walk' pluralForm='The [[count]] people went on a walk' context='male' count=myCount}}
```
oshi97, line 12: could we also get a usage with context='female', and another one where the phrase is not in the po file (to show that defaulting works properly)

src/handlebars/handlebarspreprocessor.js
```diff
@@ -90,15 +91,15 @@ class HandlebarsPreprocessor {
       return `ANSWERS.translateJS('${translatorResult}', ${parsedParams}, ${count})`;
     }
 
-    return `ANSWERS.translateJS('${translatorResult}', ${parsedParams})`;
+    const escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
```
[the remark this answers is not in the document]
cea2aj: Good catch. Adding this now

src/i18n/translator/translator.js
cea2aj: You're right! Refactoring

src/i18n/translator/translator.js
cea2aj: Ah, good point. Renaming to translationKey

tests/fixtures/handlebars/rawtemplate.hbs
cea2aj: sure

cea2aj: Added the context='female' test here and added the defaulting test to the translator tests

src/i18n/translator/translator.js
```diff
@@ -80,18 +125,39 @@ class Translator {
...
+  }
 
-    return this._i18next.t(phrase, { context, ...interpPlaceholders});
+  /**
+   * Escapes the interpolation brackets in a phrase
+   * 
+   * @param {string} phrase 
+   * @returns {string}
+   */
+  _getEscapedPhrase(phrase) {
```
alexisgrow, line 157: Could we change this name to something that describes what characters the function escapes? I really liked the name `_escapeSingleQuotes` from the other file (but ofc this one escapes brackets not quotes)

tests/handlebars/handlebarspreprocessor.js
```diff
@@ -24,18 +40,18 @@ describe('HandlebarsPreprocessor works correctly', () => {
 
   it('transpiles all "translate" and "translateJS" invocations in a JS template', () => {
     const rawJsHandlebarsContent = readFileSync(
-      path.join(__dirname, '../fixtures/handlebars/rawcomponent.js'), 'utf8');
+      path.join(__dirname, '../fixtures/handlebars/simple-example/rawcomponent.js'), 'utf8');
```
alexisgrow, line 43: For these test cases is it possible/easy to add lines to the simple fixtures containing the edge cases rather than creating entirely new tests/filters?

tests/i18n/translator/translator.js
```diff
@@ -99,20 +99,87 @@ describe('translations with one plural form (French)', () => {
...
+        'male');
+      const expectedResult = {
+        0: 'L\'homme',
+        1: 'Les hommes',
+        locale: 'fr-FR'
+      };
+      expect(translation).toEqual(expectedResult);
+    });
+
+    it('Pluralization works as expected when context = female', () => {
```
alexisgrow, line 116: Does this test provide coverage that the `Pluralization works as expected when context = male` test missing? If not, would it be possible to combine them and into something like `Pluralization works as expected with context`?

src/i18n/translator/translator.js
cea2aj: Sure, renaming to _escapeInterpolationBrackets

tests/handlebars/handlebarspreprocessor.js
cea2aj: Sure

tests/i18n/translator/translator.js
cea2aj: They cover the same code. Combining them

PR #112, opened 2020-09-02
Throw an error if the translationFile for a locale is invalid

alexisgrow:
J=SLAP-615
TEST=manual

1 - Test with a single-locale site (a site without a locale_config.json). Confirm that `jambo build` succeeds and output files look normal. 

2 - Test with a multi-locale site with out any translation files. 

For the locales in the locale_config without a "translationFile" entry, the default [locale].po fileName is used. If that file is not present, a UserError thrown, e.g.:
```
Cannot find translation file for 'es' at 'translations/es.po'
UserError: Cannot find translation file for 'es' at 'translations/es.po'
    at LocalFileParser.fetch (/Users/agrow/repo/jamboree/src/i18n/translationfetchers/localfileparser.js:38:13)
    at exports.SitesGenerator._extractTranslations (/Users/agrow/repo/jamboree/src/commands/build/sitesgenerator.js:314:10)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async exports.SitesGenerator.generate (/Users/agrow/repo/jamboree/src/commands/build/sitesgenerator.js:123:9)
```
Also verify that when there is a "translationFile", specified but the file at that path does not exist, a UserError is thrown. Manually inspect output files and check multi-lang cards to ensure that the same-language translation preprocessing (e.g. interpolation) is applied.

### DIFF
--- a/src/i18n/translationfetchers/localfileparser.js
+++ b/src/i18n/translationfetchers/localfileparser.js
@@ -1,6 +1,7 @@
 const path = require('path');
 const { readFileSync, existsSync } = require('fs');
 const { gettextToI18next } = require('i18next-conv');
+const UserError = require('../../errors/usererror');
 
 /**
  * This class parses translations from a local .PO file. The i18next-conv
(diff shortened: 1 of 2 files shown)


src/i18n/translationfetchers/localfileparser.js
```diff
@@ -31,16 +32,14 @@ class LocalFileParser {
...
-    const fileName = translationFilePath || `${locale}.po`;
-    const translationFile = path.join(this._translationsDir, fileName);
-
-    if (existsSync(translationFile)) {
-      const localeTranslations =
-        gettextToI18next(locale, readFileSync(translationFile), this._options);
-      return localeTranslations.then(data => JSON.parse(data));
+    const translationFile = path.join(this._translationsDir, translationFilePath);
+    if (!existsSync(translationFile)) {
+      throw new UserError(`Cannot find translation file for '${locale}' at '${translationFilePath}'`);
```
tmeyer2115, line 37: Just to give them a bit more detail, lets provide the full path (`translationFile`) so they can see exactly where Jambo is looking.

src/i18n/translationfetchers/localfileparser.js
```diff
@@ -31,16 +32,14 @@ class LocalFileParser {
    *                            i18next format.
   */
   async fetch(locale, translationFilePath) {
-    const fileName = translationFilePath || `${locale}.po`;
```
tmeyer2115, line 34: Do we no longer attempt a default if no `translationFile` is specified? Does this mean that the new `UserError` will be thrown if they leave out a file?

alexisgrow: Oh! My understanding was that we were removing the default. I can update to keep it if that's not expected behavior.

oshi97 (approved):

PR #113, opened 2020-09-03
Load theme translations

cea2aj:
Load both theme and custom translations and merge them

J=SLAP-614
TEST=manual

Run an internationalized jambo build and confirm that both the custom and theme translations are merged correctly

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -321,7 +321,7 @@ exports.SitesGenerator = class {
    *
    * @param {Array<string>} locales The list of locales.
    * @param {LocalizationConfig} localizationConfig
-   * @returns {Object<string, Object} A map of locale to formatted translations.
+   * @returns {Object<string>, Object} A map of locale to formatted translations.
    */
   async _extractCustomTranslations(locales, localizationConfig) {
     const localFileParser = new LocalFileParser(this.config.dirs.translations);
@@ -344,10 +344,11 @@ exports.SitesGenerator = class {
    * The translations are returned in i18next format.
    *
    * @param {Array<string>} locales The list of locales.
-   * @returns {Object<string, Object} A map of locale to formatted translations.
+   * @returns {Object<string>, Object} A map of locale to formatted translations.
    */
   async _extractThemeTranslations(locales) {
-    const themeTranslationsDir = `${this.config.dirs.themes}/${this.config.defaultTheme}/translations`
+    const themeTranslationsDir = 
+      `${this.config.dirs.themes}/${this.config.defaultTheme}/translations`;
     const localFileParser = new LocalFileParser(themeTranslationsDir);
     const translations = {};
 
@@ -363,20 +364,20 @@ exports.SitesGenerator = class {
   }
 
   /**
-   * Adds customTranslations to translations
-   * If keys confict, customTranslaitons will override them
+   * Returns an object which is a merge of customTranslations
+   * and themeTranslations. customTranslations override
+   * themeTranslations with the same translation key
    * 
-   * @param {Object<string, Object} translations 
-   * @param {Object<string, Object} customTranslations 
-   * @return {Object<string, Object}
+   * @param {Object<string>, Object} themeTranslations 
+   * @param {Object<string>, Object} customTranslations 
+   * @return {Object<string>, Object}
    */
-  _mergeTranslations(translations, customTranslations) {
-    for (const locale of Object.keys(customTranslations)) {
-      for (const [translationKey, translation] of Object.entries(customTranslations[locale]['translation'])) {
-        translations[locale]['translation'][translationKey] = translation;
-      }
-    }
-
-    return translations;
+  _mergeTranslations(themeTranslations, customTranslations) {
+    return Object.entries(customTranslations).reduce((mergedTranslations, [locale, val]) => {
+      const correspondingThemeTranslations = themeTranslations[locale] || {};
+      mergedTranslations[locale] = {};
+      mergedTranslations[locale]['translation'] = Object.assign({}, val['translation'], correspondingThemeTranslations['translation']);
+      return mergedTranslations;
+    }, {});
   }
 }
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -344,7 +344,6 @@ exports.SitesGenerator = class {
    * The translations are returned in i18next format.
    *
    * @param {Array<string>} locales The list of locales.
-   * @param {LocalizationConfig} localizationConfig
    * @returns {Object<string, Object} A map of locale to formatted translations.
    */
   async _extractThemeTranslations(locales) {
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -328,9 +328,11 @@ exports.SitesGenerator = class {
     const translations = {};
 
     for (const locale of locales) {
-      const localeTranslations = await localFileParser
-        .fetch(locale, localizationConfig.getTranslationFile(locale));
-      translations[locale] = { translation: localeTranslations };
+      if (locale !== localizationConfig.getDefaultLocale()) {
+        const localeTranslations = await localFileParser
+          .fetch(locale, localizationConfig.getTranslationFile(locale));
+        translations[locale] = { translation: localeTranslations };
+      }
     }
 
     return translations;
@@ -351,8 +353,11 @@ exports.SitesGenerator = class {
     const translations = {};
 
     for (const locale of locales) {
-      const localeTranslations = await localFileParser.fetch(locale);
-      translations[locale] = { translation: localeTranslations };
+      const translationFile = path.join(themeTranslationsDir, `${locale}.po`);
+      if (fs.existsSync(translationFile)) {
+        const localeTranslations = await localFileParser.fetch(locale);
+        translations[locale] = { translation: localeTranslations };
+      }
     }
 
     return translations;
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -309,8 +309,8 @@ exports.SitesGenerator = class {
   async _extractTranslations(locales, localizationConfig) {
     const customTranslations = await this._extractCustomTranslations(locales, localizationConfig);
     const themeTranslations = await this._extractThemeTranslations(locales);
-    const mergedTranslations = this._mergeTranslations(themeTranslations, customTranslations);
-
+    const mergedTranslations = _.merge(themeTranslations, customTranslations);
+    
     return mergedTranslations;
   }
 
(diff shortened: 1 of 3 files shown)


src/i18n/translationfetchers/localfileparser.js
```diff
@@ -35,7 +35,7 @@ class LocalFileParser {
     const fileName = translationFilePath || `${locale}.po`;
     const translationFile = path.join(this._translationsDir, fileName);
     if (!existsSync(translationFile)) {
-      throw new UserError(`Cannot find translation file for '${locale}' at '${translationFile}'`);
+      Promise.reject(new UserError(`Cannot find translation file for '${locale}' at '${translationFile}'`));
```
oshi97, line 38: I believe you want to just throw new UserError here, since the function is async it will be wrapped in a promise.

tests/i18n/translationfetchers/localfileparser.js
```diff
@@ -33,9 +33,7 @@ describe('LocalFileParser works correctly', () => {
     })
   });
 
-  it('works correctly when no translation file exists for the locale', () => {
-    return localFileParser.fetch('es').then(translations => {
-      expect(translations).toEqual({});
-    })
+  it('Rejects with an error when no translation file exists for the locale', () => {
+    expect(localFileParser.fetch('es')).rejects.toThrow();
```
oshi97, line 37: I think we want to check that the error is a UserError, otherwise the test will pass for random programming errors as well (the Promise.reject above is not actually returned, so I think what's currently being caught is a separate error)

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+  /**
+   * Parses the local translation files for the provided locales. 
+   * Parses translations in the theme translations folder
+   * The translations are returned in i18next format.
+   *
+   * @param {Array<string>} locales The list of locales.
+   * @returns {Object<string, Object} A map of locale to formatted translations.
+   */
+  async _extractThemeTranslations(locales) {
+    const themeTranslationsDir = `${this.config.dirs.themes}/${this.config.defaultTheme}/translations`
```
tmeyer2115, line 350: LL, put this on the next line.

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+        translations[locale] = { translation: localeTranslations };
+      }
+    }
+
+    return translations;
+  }
+
+  /**
+   * Adds customTranslations to translations
+   * If keys confict, customTranslaitons will override them
```
oshi97, line 367: small typo

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+   * 
+   * @param {Object<string, Object} translations 
+   * @param {Object<string, Object} customTranslations
```
oshi97, line 370: missing closing angle brackets

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+    return translations;
+  }
+
+  /**
+   * Parses the local translation files for the provided locales. 
+   * Parses translations in the theme translations folder
+   * The translations are returned in i18next format.
+   *
+   * @param {Array<string>} locales The list of locales.
+   * @returns {Object<string, Object} A map of locale to formatted translations.
```
oshi97, line 347: missing closing angle bracket

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+  /**
+   * Adds customTranslations to translations
+   * If keys confict, customTranslaitons will override them
+   * 
+   * @param {Object<string, Object} translations 
+   * @param {Object<string, Object} customTranslations 
+   * @return {Object<string, Object}
+   */
+  _mergeTranslations(translations, customTranslations) {
+    for (const locale of Object.keys(customTranslations)) {
```
tmeyer2115, line 374: In general, I feel we should avoid mutating input arguments like `translations`. Internal mutation in a method can cause confusion and add complexity elsewhere. Other consumers need to anticipate that this method alters its input.

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+   * The translations are returned in i18next format.
    *
    * @param {Array<string>} locales The list of locales.
    * @param {LocalizationConfig} localizationConfig
    * @returns {Object<string, Object} A map of locale to formatted translations.
    */
   async _extractTranslations(locales, localizationConfig) {
+    const customTranslations = await this._extractCustomTranslations(locales, localizationConfig);
+    const themeTranslations = await this._extractThemeTranslations(locales);
+    const mergedTranslations = this._mergeTranslations(themeTranslations, customTranslations);
```
oshi97, line 312: oo maybe in the future we want to do some kind of namespacing thing so different theme translations dont conflict, but this is definitely good for now!

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+    }
+
+    return translations;
+  }
+
+  /**
+   * Adds customTranslations to translations
+   * If keys confict, customTranslaitons will override them
+   * 
+   * @param {Object<string, Object} translations
```
tmeyer2115, line 369: themeTranslations for extra clarity?

src/commands/build/sitesgenerator.js
```diff
@@ -298,21 +298,83 @@ exports.SitesGenerator = class {
...
+   * If keys confict, customTranslaitons will override them
+   * 
+   * @param {Object<string, Object} translations 
+   * @param {Object<string, Object} customTranslations 
+   * @return {Object<string, Object}
+   */
+  _mergeTranslations(translations, customTranslations) {
+    for (const locale of Object.keys(customTranslations)) {
+      for (const [translationKey, translation] of Object.entries(customTranslations[locale]['translation'])) {
+        translations[locale]['translation'][translationKey] = translation;
```
tmeyer2115, line 376: I think there's a more functional way to create the merged translations. We typically prefer to do things like this in a more functional style. Something like this might work:

return Object.entries(translations).reduce((mergedTranslations, [locale, val]) => {
  const customTranslations = customTranslations[locale] || {};
  mergedTranslations[locale]['translation'] = Object.assign({}, val['translation'], customTranslations['translation']);
  return mergedTranslations;
}, {});

src/i18n/translationfetchers/localfileparser.js
cea2aj: oh, makes sense!

tests/i18n/translationfetchers/localfileparser.js
cea2aj: Updating. It looks like it is actually catching the user error though

src/commands/build/sitesgenerator.js
cea2aj: Ya, I was debating whether or not to do this. I went with mutating it to avoid the computation of copying all the translations, however I will change it!

src/commands/build/sitesgenerator.js
cea2aj: sure

src/commands/build/sitesgenerator.js
cea2aj: Sure, I'll try that out

src/commands/build/sitesgenerator.js
tmeyer2115: I don't think the computation cost will be too noticeable, given the size of map we expect to deal with. So, I'd argue that the complexity from having mutated arguments isn't worth it. If we find this becomes a bottleneck eventually, we can reassess!

src/commands/build/sitesgenerator.js
cea2aj: I had to modify this a little because we want to include custom translations that may not have corresponding theme translations of the same locale and vice versa

src/commands/build/sitesgenerator.js
```diff
@@ -298,23 +298,86 @@ exports.SitesGenerator = class {
...
+    return translations;
+  }
+
+  /**
+   * Parses the local translation files for the provided locales. 
+   * Parses translations in the theme translations folder
+   * The translations are returned in i18next format.
+   *
+   * @param {Array<string>} locales The list of locales.
+   * @returns {Object<string>, Object} A map of locale to formatted translations.
```
alexisgrow, line 347: Should the return type be "Object<string, Object>"?

cea2aj: Oh, you're right. I'll change that real quick

src/commands/build/sitesgenerator.js
```diff
@@ -298,23 +298,86 @@ exports.SitesGenerator = class {
...
+   */
+  async _extractCustomTranslations(locales, localizationConfig) {
     const localFileParser = new LocalFileParser(this.config.dirs.translations);
     const translations = {};
 
     for (const locale of locales) {
-      const localeTranslations = await localFileParser
-        .fetch(locale, localizationConfig.getTranslationFile(locale));
-      translations[locale] = { translation: localeTranslations };
+      if (locale !== localizationConfig.getDefaultLocale()) {
```
alexisgrow, line 331: I was thinking about this earlier, are we safe to assume that the site is always written in the language of the default locale?

src/commands/build/sitesgenerator.js
```diff
@@ -298,23 +298,86 @@ exports.SitesGenerator = class {
...
+   * 
+   * @param {Object<string, Object>} themeTranslations 
+   * @param {Object<string, Object>} customTranslations 
+   * @return {Object<string, Object>}
+   */
+  _mergeTranslations(themeTranslations, customTranslations) {
+    return Object.entries(customTranslations).reduce((mergedTranslations, [locale, val]) => {
+      const correspondingThemeTranslations = themeTranslations[locale] || {};
+      mergedTranslations[locale] = {};
+      mergedTranslations[locale]['translation'] = Object.assign({}, val['translation'], correspondingThemeTranslations['translation']);
```
alexisgrow, line 379: Line length here and also potentially on line 376 depending on what our cut-off is for this repo! Other than that, pr LGTM!

src/commands/build/sitesgenerator.js
cea2aj: I think that's reasonable. This if statement was just added to prevent the local file parser from trying to load a PO file for default locale which doesn't need translations

src/commands/build/sitesgenerator.js
```diff
@@ -298,23 +298,86 @@ exports.SitesGenerator = class {
...
+   * Returns an object which is a merge of customTranslations
+   * and themeTranslations. customTranslations override
+   * themeTranslations with the same translation key
+   * 
+   * @param {Object<string, Object>} themeTranslations 
+   * @param {Object<string, Object>} customTranslations 
+   * @return {Object<string, Object>}
+   */
+  _mergeTranslations(themeTranslations, customTranslations) {
+    return Object.entries(customTranslations).reduce((mergedTranslations, [locale, val]) => {
```
tmeyer2115, line 376: Thinking a bit more, I'm not sure that we should iterate over the `customTranslations` locales like this. It's entirely possible that the Theme has translations for a locale that has no custom translations yet. In this case, we'd erroneously report no merged translations for the locale. The converse is possible as well (where there are custom translations for a locale that has no Theme translations). Because of this, we may want to iterate over the set of locales created from combining the keys of `themeTranslations` with the keys of `customTranslations`.

oshi97, line 376: Actually, would it be possible to use lodash.merge here? (Sorry I know you already rewrote this function once).

src/commands/build/sitesgenerator.js
cea2aj: Alternatively we could make a variable somewhere that represents the locale which doesn't need translations

src/commands/build/sitesgenerator.js
oshi97: Seeing Tom's comment above about having to iterate over the set of both object's keys makes me doubly think it would be nice to just use lodash.merge

cea2aj: Ya, i'll try that looks like the functionality we're going for

cea2aj: lodash.merge does the trick here!

tmeyer2115 (approved):

PR #115, opened 2020-09-08
Add Javascript linting to Jambo.

tmeyer2115:
This PR adds ESLint to Jambo, a linting framework for Javascript. This will
help make sure that any code added to Jambo adheres to our style guidelines
and best practices. For now, I've enabled only a couple of rules:

- All quotes must use be wrapped in single quotes.
- Named functions do not have a space between function name and the opening
  '('.

As our style evolves, we will codify it by adding more rules to ESLint. In
my next PR, I will ensure that the Jambo codebase adheres to these rules and
that ESLint is run as part of 'npm run test'.

TEST=none

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,18 @@
+{
+    "env": {
+        "commonjs": true,
+        "es2021": true,
+        "node": true,
+        "jest": true
+    },
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "ignorePatterns": ["tests/fixtures/**/*.js"],
+    "rules": {
+        "quotes": ["error", "single"],
+        "space-before-function-paren": ["error", {
+            "named": "never"
+        }]
+    }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -3202,6 +3202,53 @@
         "minimist": "^1.2.0"
       }
     },
+    "@eslint/eslintrc": {
+      "version": "0.1.3",
+      "resolved": "https://registry.npmjs.org/@eslint/eslintrc/-/eslintrc-0.1.3.tgz",
+      "integrity": "sha512-4YVwPkANLeNtRjMekzux1ci8hIaH5eGKktGqR0d3LWsKNn5B2X/1Z6Trxy7jQXl9EBGE6Yj02O+t09FMeRllaA==",
+      "dev": true,
+      "requires": {
+        "ajv": "^6.12.4",
+        "debug": "^4.1.1",
+        "espree": "^7.3.0",
+        "globals": "^12.1.0",
+        "ignore": "^4.0.6",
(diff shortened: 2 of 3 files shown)


.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+        "jest": true
+    },
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "ignorePatterns": ["tests/fixtures/**/*.js"],
+    "rules": {
+        "quotes": ["error", "single"],
+        "space-before-function-paren": ["error", {
+            "named": "never"
```
alexisgrow, line 15: I think this rule conflicts with one of our rules for the Answers JS lib's linting config. Would it be possible to use the same ruleset for both? 

I don't really have a preference the spacing between function name and parenthesis, but it would be really nice if the rules were consistent so we don't have to swap linters per-repo.

tmeyer2115: Yeah, this does conflict with the SDK. I'm not a big fan of `semistandard` and the conventions it enforces in the SDK. I'd like us to move over to ESLint there as well and we'd definitely want to have the same core rule set for both. If people object to having two different linters in the interim, I can hold off until we do both!

tmeyer2115: `semistandard` and `standard` are really rigid and prescriptive (which is kind of their point). They don't let you easily change or configure rules. I think it's important to have this flexibility. 
https://github.com/standard/standard#i-disagree-with-rule-x-can-you-change-it

cea2aj:
I'm a fan! I've felt before that adding a linter would benefit Jambo

.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+        "es2021": true,
+        "node": true,
+        "jest": true
+    },
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "ignorePatterns": ["tests/fixtures/**/*.js"],
+    "rules": {
+        "quotes": ["error", "single"],
```
oshi97, line 13: Could we also add quote-props? the default is `always` but I think I prefer as-needed or consistent-as-needed

.eslintrc.json
alexisgrow: Ah amazing! That sounds good to me, no objections here!

.eslintrc.json
tmeyer2115: Yep! I can add that. Okay if I do this in my next PR? I have a couple other rules to add as well. I wanted to put up this PR first, to make sure we were all on board. Next PR will be adding a couple others rules and fixing all violations in the code base.

oshi97: yeah definitely on board! definitely prefer eslint over semistandard its way more mainstream

oshi97 (approved):

PR #116, opened 2020-09-08
Escape double quotes for pluralized translations

cea2aj:
Escape double quotes when transpiling translateJS and translate calls

When JSON is represented as a string literal between single quotes, double quotes need to be escaped twice and single quotes need to be escaped once. This PR adds the proper escapement to the double quotes. Previously double quotes weren't escaped at all would cause jambo to produce invalid JSON when a translation included a double quote.

This change also causes the transpiled calls to also escape all the double quotes surrounding JSON keys and values. While the additional escapes may seem unnecessary, they are still valid.  Previously these escapes were being created inside calls to ANSWERS.registerTemplate() because templates were double-quoted strings. With these new changes, the double quote escapes surrounding JSON keys and values inside ANSWERS.registerTemplate() calls are still the same. The only new changes are the extra escapes before double quotes within pluralized translations strings.

J=SLAP-636
TESt=manual, auto

Add new tests that cover the escaped double quotes. Also add more intermixed markdown tests.
Manually tests an internationalized jambo site and observe pluralized translations containing double quotes working that were previously broken.


src/handlebars/handlebarspreprocessor.js
```diff
@@ -108,29 +115,44 @@ class HandlebarsPreprocessor {
...
    */
-  _createRuntimeCallForHBS(translatorResult, interpolationParams) {
+  _createRuntimeCallForHBS(translatorResult, interpolationParams, needsPluralization) {
     const paramsString = Object.entries(interpolationParams)
       .reduce((params, [paramName, paramValue]) => {
         return params + `${paramName}=${paramValue} `;
       }, '');
 
-    const escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
+    let escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
```
oshi97, line 127: I did some googling and you might just be able to call JSON.stringify on this (again) to escape stuff? https://gist.github.com/mathiasbynens/d6e10171d44a59bb5664617c64ff2763

I did a test with 
```js 
const a = `applebee's homemade pie's! applebees.com/sdf\\12`
```
and also with the gist's example and it looks like it could work?

src/handlebars/handlebarspreprocessor.js
```diff
@@ -81,22 +87,23 @@ class HandlebarsPreprocessor {
...
     let parsedParams = JSON.stringify(interpolationParams);
     parsedParams = parsedParams.replace(/[\'\"]/g, '');
 
+    let escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
+
     if (needsPluralization) {
       const count = interpolationParams.count;
-      return `ANSWERS.translateJS('${translatorResult}', ${parsedParams}, ${count})`;
+      escapedTranslatorResult = this._escapeDoubleQuotes(escapedTranslatorResult);
+      return `ANSWERS.translateJS('${escapedTranslatorResult}', ${parsedParams}, ${count})`;
```
alexisgrow, line 104: Do you know if it's necessary to wrap the escapedTranslatorResult in a string here? In the SDK we don't, e.g. here's an example of a transpiled value with interpolation + pluralization:
```js
var test1 = ANSWERS.translateJS({"0":"Il y a [[number]] article [[name]]","1":"Il y a [[number]] articles [[name]]","locale":"fr-FR"}, {name:this._config.title,number:this._config.details.length}, 2);

```

tmeyer2115 (commented):
What if we were to not use JSON serialization at all? When creating calls to `ANSWERS.translateJS`, we'd just pass an actual object. That'd be nice because we wouldn't have to serialize just to deserialize in `translateJS`. When creating calls to `runtimeTranslation`, we'd pass named arguments for each entry in the object. For example:

`{ phrase: 'Blah blah', pluralForm1: 'blahs blahs' pluralForm2: 'blah blahs' }` 

would become:

`{ runtimeTranslation phrase='Blah blah' pluralForm1='blahs blahs' pluralForm2='blah blahs' }`.

Would this help the situation at all? As we discussed yesterday, I'm worried by the complexity of the escaping logic we have sprinkled around the codebase. I think that with the approach outline above, we'd just need to escape single quotes (since we wrap everything in them).

src/handlebars/handlebarspreprocessor.js
cea2aj: I'll give that a try! Although I'm think we might want to go with Tom's suggestion of not using JSON at all and simply passing all the object entries as named arguments to the translator

src/handlebars/handlebarspreprocessor.js
cea2aj: ANSWERS.translateJS() expects a string or stringified json as the first parameter. Have you confirmed that that translation works?

cea2aj:
> What if we were to not use JSON serialization at all? When creating calls to `ANSWERS.translateJS`, we'd just pass an actual object. That'd be nice because we wouldn't have to serialize just to deserialize in `translateJS`. When creating calls to `runtimeTranslation`, we'd pass named arguments for each entry in the object. For example:
> 
> `{ phrase: 'Blah blah', pluralForm1: 'blahs blahs' pluralForm2: 'blah blahs' }`
> 
> would become:
> 
> `{ runtimeTranslation phrase='Blah blah' pluralForm1='blahs blahs' pluralForm2='blah blahs' }`.
> 
> Would this help the situation at all? As we discussed yesterday, I'm worried by the complexity of the escaping logic we have sprinkled around the codebase. I think that with the approach outline above, we'd just need to escape single quotes (since we wrap everything in them).

I think that makes a lot of sense. It's simpler and less error prone. Currently the SDK translator attempts a JSON parse on the phrase parameter. If it's successful, it assumes a pluralization is taking place. If it fails, it assumes it is translating a single string. It works, but it requires that we always send it valid JSON, which seems difficult to always get right as a string literal inside a HBS template. I'm going to try out Oliver's suggestion of the double stringify and see how robust that is. If that doesn't work, I think we'll want to make this change.

src/handlebars/handlebarspreprocessor.js
cea2aj: This almost did the trick, however the resulting string is surrounded by escaped double quotes rather than single quotes. I'm going to talk to Tom about making this change: https://github.com/yext/jambo/pull/116#pullrequestreview-484980929

cea2aj:
Spoke with @tmeyer2115 offline. He said that we can get this fix in now and we can revisit our JSON serialization and escaping with another item.

tmeyer2115 (approved):

PR #117, opened 2020-09-09
Add additional linting rules and ensure codebase compliance.

tmeyer2115:
This PR adds a couple of new rules to the ESLint config:

- 'max-len' enforces a line length of 90 characters (I've excluded 'require'
  statements).
- 'quote-props' ensures we don't wrap object keys in quotes unless
  absolutely necessary.

I have also included the changes needed to make the code compliant with all
of the ESLint rules. You can automatically fix some linting errors with
'eslint -fix'. Lastly, linting is now run as part of 'npm run test'.

TEST=auto

Made sure 'npm run test' did not report any issues.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,7 +12,13 @@
     "rules": {
         "quotes": ["error", "single"],
         "space-before-function-paren": ["error", {
-            "named": "never"
+            "named": "never",
+            "anonymous": "never"
+        }],
+        "quote-props": ["error", "as-needed"],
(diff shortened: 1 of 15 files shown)


cea2aj (approved):
LGTM! When we merge in localized-pages I don't think we will be able to stay within the 90 character line length with our translation helper calls since we don't support multi-line translation calls. Those should only show up in tests though, so I don't think it will be a problem. Just something to keep in mind.

PR #118, opened 2020-09-10
Add escapeHTML flag to the HBS translation helper

cea2aj:
Add escapeHTML flag to the HBS translation helper

When escapeHTML=false, runtimeTranslation calls are wrapped in triple curly braces. This allows translations with HTML to render as HTML. When escapeHTML=true, runtimeTranslation calls are wrapped in double curly braces which causes handlebars to escape HTML and display it as a string. For static translations, the escapeHTML flag determines whether or not Handlebars.Utils.escapeExpression() should be used.

J=SLAP-637
TEST=auto, manual

Add unit tests with escapeHTML flags that confirm the number of braces surrounding runtimeTranslation calls. For static translations, confirm that Handlebars.Utils.escapeExpression() is used when escapeHTML=true. On an internationalized answers site, observe proper escapement based on the escapeHTML flag.


cea2aj:
What do you all think of the name of the flag "escapeHTML"? I was also considering "allowHTML". @tmeyer2115 also suggested "shouldUnescape".

src/handlebars/handlebarspreprocessor.js
```diff
@@ -130,7 +140,10 @@ class HandlebarsPreprocessor {
       escapedTranslatorResult = this._escapeDoubleQuotes(escapedTranslatorResult);
     }
 
-    return `{{ runtimeTranslation phrase='${escapedTranslatorResult}' ${paramsString}}}`;
+    if (shouldEscapeHTML) {
```
tmeyer2115, line 143: Minor nit, can we use the ternary operator here?

src/handlebars/handlebarspreprocessor.js
```diff
@@ -62,11 +63,16 @@ class HandlebarsPreprocessor {
...
-      return invocation.getInvokedHelper() === 'translateJS' ? 
-        `'${escapedTranslatorResult}'`: 
-        translatorResult;
+      if (invocation.getInvokedHelper() === 'translateJS' ) {
+        const escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
+        return `'${escapedTranslatorResult}'`;
+      }
+
+      if (invocation.shouldEscapeHTML()) {
+        return Handlebars.Utils.escapeExpression(translatorResult);
```
tmeyer2115, line 72: It's been a bit, so I don't remember exactly. Do we need this and to wrap it in double curly braces? Is the latter not sufficient?

src/handlebars/handlebarspreprocessor.js
cea2aj: Sure

src/handlebars/handlebarspreprocessor.js
cea2aj: The double curly braces don't apply here since this is a static translation. Previously, the translatorResult was getting put right on the page during transpilation which included unescaped HTML. You can see an example in the old tests if you CTRL-F for "<button>L'homme</button>" on the GitHub "Files Changed" tab

tmeyer2115: Ahh, that's right!

tmeyer2115 (approved):

PR #119, opened 2020-09-10
Allow 'apiKey' only in the 'global_config' file.

tmeyer2115:
Previously, we assumed that each locale in the 'locale_config' would need
the ability to specify a localized 'apiKey'. This assumption was not correct.
The 'apiKey' is the same for all experiences in a business. There will be
different experiences per locale, but the 'apiKey' is constant. Because it
is something that is business-specific, we will now only allow it in the
'global_config' file. Any 'apiKey' in 'locale_config' will be ignored.

J=SLAP-640
TEST=auto, manual

Fixed and ran unit tests. Created a site with English and French page sets.
Set the 'apiKey' only in the 'global_config' and saw it used for both locales.

### DIFF
--- a/src/commands/build/globalconfiglocalizer.js
+++ b/src/commands/build/globalconfiglocalizer.js
@@ -20,13 +20,10 @@ module.exports = class GlobalConfigLocalizer {
   localize(globalConfig, locale) {
     const experienceKey = this._localizationConfig.getExperienceKey(locale)
       || globalConfig.getExperienceKey();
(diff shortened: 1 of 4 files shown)


alexisgrow (approved):
LGTM!

cea2aj (changes requested):
Let me verify the build validation step real quick before you merge

cea2aj (approved):
Looks good

PR #121, opened 2020-09-10
Add an SDK bundle locale override option to 'locale_config'

alexisgrow:
Add support for an "sdkBundleLocaleOverride" parameter in the individual locale objects in the locale_config. Jambo then adds a parameter "sdkBundleLocale" to the object for the Handlebars templates, which is the version of the SDK bundle that should be used. 

Right now this PR introduces fallback logic that could arguably belong in the theme -- the `sdkBundleLocale = sdkBundleLocaleOverride || locale || 'en';` We can absolutely move it in there, but it's just so messy to do that kind of fallback logic in-line in Handlebars I put it here for now. As a visual, this is how the theme call looks with the existing Jambo code:
```hbs
<script src="https://assets.sitescdn.net/answers/v{{global_config.sdkVersion}}/{{#unless (eq sdkBundleLocale 'en')}}{{sdkBundleLocale}}-{{/unless}}answerstemplates.compiled.min.js"></script>
```
And if we moved the logic to the theme, this is how it would look: 
```hbs
<script src="https://assets.sitescdn.net/answers/v{{global_config.sdkVersion}}/{{#if sdkBundleLocale}}{{#unless (eq sdkBundleLocale 'en')}}{{sdkBundleLocale}}-{{/unless}}{{else if global_config.locale}}{{#unless (eq global_config.locale 'en')}}{{global_config.locale}}-{{/unless}}answerstemplates.compiled.min.js"></script>
```

TEST=manual
J=SLAP-633

Tested with the changes from PR #120 (https://github.com/yext/jambo/pull/120) and the changes in the theme here: https://github.com/yext/answers-hitchhiker-theme/pull/317. Viewed output .html files for English (no bundle prefix), Spanish (es bundle prefix), and French (Belgium), fr-be (used override here for 'fr') and verified output was as expected.

### DIFF
--- a/src/commands/build/pagesetsbuilder.js
+++ b/src/commands/build/pagesetsbuilder.js
@@ -55,11 +55,16 @@ module.exports = class PageSetsBuilder {
       const localizedGlobalConfig = new GlobalConfigLocalizer(this._localizationConfig)
         .localize(this._globalConfig, locale);
 
+      const sdkBundleLocale = this._localizationConfig.getSdkBundleLocaleOverride(locale)
+        || locale
+        || 'en';
+
       pageSets.push(new PageSet({
         locale: locale,
         pages: this._buildPages(pageConfigs, templates),
         globalConfig: localizedGlobalConfig,
-        params: this._localizationConfig.getParams(locale)
+        params: this._localizationConfig.getParams(locale),
+        sdkBundleLocale: sdkBundleLocale
       }));
     }
     return pageSets;
--- a/src/commands/build/pagewriter.js
+++ b/src/commands/build/pagewriter.js
@@ -43,13 +43,10 @@ module.exports = class PageWriter {
       }
 
       console.log(`Writing output file for the '${page.getName()}' page`);
-      const templateArguments = this._buildArgsForTemplate({
-        pageConfig: page.getConfig(),
-        relativePath: this._calculateRelativePath(page.getOutputPath()),
-        params: pageSet.getParams(),
-        globalConfig: pageSet.getGlobalConfig().getConfig(),
-        pageNameToConfig: pageSet.getPageNameToConfig(),
-      });
+      const templateArguments = this._buildArgsForTemplate(
+        page.getConfig(),
+        this._calculateRelativePath(page.getOutputPath()),
+        pageSet);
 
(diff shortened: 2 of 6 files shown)


tmeyer2115 (commented):
I've thought about this issue a bit more. With our upcoming CBD work, we will be ripping HH specific things out of Jambo and adding them to the Theme. So, I think we should avoid adding to this work. I think we can accomplish this functionality purely with Theme changes. Here's what we'd do:

- We add `sdkBundleLocaleOverride` to the `params` object for a locale. This keeps things generic for Jambo. It doesn't have to know anything about this particular param, it already makes the contents of `params` available during HBS compilation.
- We do add the ugly defaulting logic to the Theme. This isn't ideal, but we have plans to allow themes to register their own HBS helpers with Jambo. We can refactor with this down the road to clean this logic up in the Theme.

I know that adding this logic to Jambo would allow us to keep the Theme cleaner. But, I think we need to start enforcing that boundary between Jambo and Theme. This is Theme-specific logic.

alexisgrow:
> I've thought about this issue a bit more. With our upcoming CBD work, we will be ripping HH specific things out of Jambo and adding them to the Theme. So, I think we should avoid adding to this work. I think we can accomplish this functionality purely with Theme changes. Here's what we'd do:
> 
> * We add `sdkBundleLocaleOverride` to the `params` object for a locale. This keeps things generic for Jambo. It doesn't have to know anything about this particular param, it already makes the contents of `params` available during HBS compilation.
> * We do add the ugly defaulting logic to the Theme. This isn't ideal, but we have plans to allow themes to register their own HBS helpers with Jambo. We can refactor with this down the road to clean this logic up in the Theme.
> 
> I know that adding this logic to Jambo would allow us to keep the Theme cleaner. But, I think we need to start enforcing that boundary between Jambo and Theme. This is Theme-specific logic.

Totally agree - it seems odd that this would be a first-class thing in Jambo. I'll add the defaulting logic to the theme expecting this value from the params object. As part of that theme PR, it may also be nice to separate the answers bundle names from the rest of scripts/core.hbs, in case the HH's would like to shadow the bundle selecting logic but keep the rest of core.hbs as-is.

PR #123, opened 2020-09-17
Add jambo matches Handlebars helper

alexisgrow:
TEST=manual
J=SLAP-626

Use helper in a jambo site, test adding param to the global_config: `"alexis-text-param": "static/assets/image.png"` then putting this snippet in the Handlebars.
```hbs
{{#if (matches global_config.alexis-text-param 'static\/assets\/[^"]*' )}}
    works!
{{/if}}
```
Inspect output HTML and see "works!" in the file. Change "if" to "unless" and see that "works!" is no longer in the output file.

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -303,7 +303,7 @@ exports.SitesGenerator = class {
       return (prefix + id);
     });
 
-    hbs.registerHelper('contains', function(str, regexPattern) {
+    hbs.registerHelper('matches', function(str, regexPattern) {
       const regex = new RegExp(regexPattern);
       return str.match(regex);
     });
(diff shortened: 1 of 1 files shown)


src/commands/build/sitesgenerator.js
```diff
@@ -303,6 +303,11 @@ exports.SitesGenerator = class {
       return (prefix + id);
     });
 
+    hbs.registerHelper('contains', function(str, regexPattern) {
```
tmeyer2115, line 306: What would you think about calling this 'matches' instead of 'contains'? That seems more inline with what this method does, it checks if a string matches the provided regex.

alexisgrow: Lol yes I totally agree. That's what I originally had here (as you can see from my edit of the PR comment) but I changed it at the last minute. I'll swap it back to "matches"!

tmeyer2115 (approved):

PR #127, opened 2020-09-21
Add null check to matches handlebars helper

alexisgrow:
Currently if we try to use the "matches" helper with a variable that does not exist, it fails. For example:
```hbs
{{#if (matches thisVariableDoesNotExist 'test')}}{{/if}}
```
throws an
```
TypeError: Cannot read property 'match' of undefined
    at Object.<anonymous> 
    ...
```
Error. 

TEST=manual
J=SPR-667

Use handlebars helper with variable that does not exist. Confirm it returns false (rather than causing the jambo build to fail)


cea2aj (approved):
🔥

oshi97 (approved):
lgtm!

PR #128, opened 2020-09-21
Update output of transpilation calls

cea2aj:
Update the transpilation of runtime calls calls to no longer create stringified json

For calls to ANSWERS.processTranslation(), pass an object as the first parameter rather than stringified JSON. For calls to the processTranslation HBS helper, instead of passing pluralizations through the phrase param, pass pluralForms as their own parameters. This simplifies our escaping logic to only require the escaping of single quotes. Also, no longer create a locale parameter when creating transpiled calls.

J=SLAP-638
TEST=auto, manual

Update unit tests and confirm they pass. Run an internationalized jambo build with the corresponding branch in the SDK and verify that the translations are correct. Also inspect the HTML generated through the jambo build to verify that the calls to the helpers are correct.

### DIFF
--- a/src/handlebars/handlebarspreprocessor.js
+++ b/src/handlebars/handlebarspreprocessor.js
@@ -141,7 +141,11 @@ class HandlebarsPreprocessor {
    * @returns {string} The call to the 'processTranslation' helper.
    */
   _createRuntimeCallForHBS(
-  translatorResult, interpolationParams, needsPluralization, shouldEscapeHTML) {
+    translatorResult, 
+    interpolationParams, 
+    needsPluralization, 
+    shouldEscapeHTML) 
+  {
     const translationString = needsPluralization ?
       Object.entries(translatorResult)
         .reduce((params, [paramName, paramValue]) => {
--- a/src/handlebars/handlebarspreprocessor.js
+++ b/src/handlebars/handlebarspreprocessor.js
@@ -106,8 +106,25 @@ class HandlebarsPreprocessor {
 
     if (needsPluralization) {
       const count = interpolationParams.count;
+      const pluralFormPairs = this._getPluralFormPairs(translatorResult);
 
-      const translationPairs = Object.entries(translatorResult)
+      return `ANSWERS.processTranslation(${pluralFormPairs}, ${parsedParams}, ${count})`;
+    }
+    const escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
+
+    return `ANSWERS.processTranslation('${escapedTranslatorResult}', ${parsedParams})`;
+  }
+
+  /**
+   * Constructs a string representation of a translatorResult. This output is similar to
+   * JSON.stringiy(), however keys are not surrounded by quotes, and values are
+   * surrounded by single quotes.
+   * 
+   * @param {Object<number,string>} translatorResult 
+   * @returns {string}
+   */
+  _getPluralFormPairs(translatorResult) {
+    const pluralFormPairs = Object.entries(translatorResult)
         .reduce((params, [pluralFormIndex, pluralForm], index, array) => {
           const escapedPluralForm = this._escapeSingleQuotes(pluralForm);
           const accumulatedParams = params + `${pluralFormIndex}:'${escapedPluralForm}'`;
@@ -118,12 +135,7 @@ class HandlebarsPreprocessor {
             accumulatedParams + ',';
         }, '');
 
-      const translations = '{' + translationPairs + '}';
-
-      return `ANSWERS.processTranslation(${translations}, ${parsedParams}, ${count})`;
-    }
-    const escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
-    return `ANSWERS.processTranslation('${escapedTranslatorResult}', ${parsedParams})`;
+      return '{' + pluralFormPairs + '}';
   }
 
   /**
@@ -132,7 +144,7 @@ class HandlebarsPreprocessor {
    * interpolation parameters.
    *
    * @param {Object|string} translatorResult The translation(s) for the phrase.
-   * @param {Object<string, ?>} interpolationParams The needed interpolation parameters
+   * @param {Object<string, ?>} interpolationValues The needed interpolation parameters
    *                                                (including 'count').
    * @param {boolean} needsPluralization If pluralization is required when translating.
    * @param {boolean} shouldEscapeHTML If HTML should be escaped. If false, wrap the call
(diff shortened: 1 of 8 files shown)


src/handlebars/handlebarspreprocessor.js
```diff
@@ -128,25 +138,28 @@ class HandlebarsPreprocessor {
    * @param {boolean} shouldEscapeHTML If HTML should be escaped. If false, wrap the call
    *                                   in triple curly braces. If true, wrap in in double
    *                                   double curly braces.
-   * @returns {string} The string-ified call to the 'processTranslation' helper.
+   * @returns {string} The call to the 'processTranslation' helper.
    */
   _createRuntimeCallForHBS(
-    translatorResult, interpolationParams, needsPluralization, shouldEscapeHTML)
-  {
+  translatorResult, interpolationParams, needsPluralization, shouldEscapeHTML) {
```
cea2aj, line 144: I put these params on a new line since the linter was throwing an error do to the line length. Is this how we want these params to look?

tests/handlebars/handlebarspreprocessor.js
```diff
@@ -112,36 +114,4 @@ describe('HandlebarsPreprocessor works correctly', () => {
...
-      expect(handlebarsPreprocessor.process(raw)).toEqual(processed);
-    });
-
-    it('transpiles commented out "translate" invocations correctly', () => {
-      const raw = '{{!-- {{ translate phrase=\'singular\' pluralForm=\'plural\' }} --}}';
-      const processed =
-        '{{!-- {{ processTranslation phrase=\'{\\"0\\":\\"singular\\",\\"1\\":' +
-        '\\"plural\\",\\"locale\\":\\"en\\"}\' }} --}}';
-      expect(handlebarsPreprocessor.process(raw)).toEqual(processed);
-    });
```
cea2aj, line 145: I got rid of these tests because the first test was already being covered and I put the second test in the fixtures.

src/handlebars/handlebarspreprocessor.js
oshi97: *thinking* beep beep @tmeyer2115

tmeyer2115: In Java world, if not all params can fit on the same line for a method declaration, we do:
methodName(
-> param1,
-> param2,
...
-> paramN)
{
.....

I'm kinda partial to that. But, I'm open to other suggestions if people have different preferences.

cea2aj: Sure, that's definitely an improvement

src/handlebars/handlebarspreprocessor.js
```diff
@@ -128,25 +138,32 @@ class HandlebarsPreprocessor {
...
+    interpolationParams, 
+    needsPluralization, 
+    shouldEscapeHTML) 
   {
+    const translationString = needsPluralization ?
+      Object.entries(translatorResult)
+        .reduce((params, [paramName, paramValue]) => {
+          paramValue = this._escapeSingleQuotes(paramValue);
+          return paramName === 'locale' ?
+            params + `locale='${paramValue}' ` :
```
oshi97, line 154: I kind of wish translator returned something like 
```js
{
pluralForms: ['blah', 'plural blah'],
locale: 'bl'
}
```
or maybe just not including locale in the plural information anymore, since the sdk is now locale aware? I'm okay with what we're doing here even though it's not the most elegant, just some thoughts

src/handlebars/handlebarspreprocessor.js
```diff
@@ -98,21 +98,31 @@ class HandlebarsPreprocessor {
...
-    let escapedTranslatorResult = this._escapeSingleQuotes(translatorResult);
-
     if (needsPluralization) {
       const count = interpolationParams.count;
-      escapedTranslatorResult = this._escapeDoubleQuotes(escapedTranslatorResult);
-      return 'ANSWERS.processTranslation(' +
-        `'${escapedTranslatorResult}', ${parsedParams}, ${count})`;
-    }
 
+      const translationPairs = Object.entries(translatorResult)
```
tmeyer2115, line 110: Would something like `pluralFormPairs` be more descriptive here?

src/handlebars/handlebarspreprocessor.js
```diff
@@ -128,25 +138,32 @@ class HandlebarsPreprocessor {
...
+   * @returns {string} The call to the 'processTranslation' helper.
    */
   _createRuntimeCallForHBS(
-    translatorResult, interpolationParams, needsPluralization, shouldEscapeHTML)
+    translatorResult, 
+    interpolationParams, 
+    needsPluralization, 
+    shouldEscapeHTML) 
   {
+    const translationString = needsPluralization ?
```
tmeyer2115, line 149: What would you think about renaming this as well? It's a small nit, but this can hold more than one translation. It's also not just a translation(s), represented as a string. A key is contained in the string as well.

src/handlebars/handlebarspreprocessor.js
tmeyer2115: I think we should chat with Product about this. At the very least, it seems like the run-time methods should default the locale to the SDK's locale. That way, 99% of people don't need to specify it. Maybe we give people the ability to override. Or, Product says no need, and we remove locale from the SDK `processTranslation` methods altogether.

tmeyer2115: Not needing to pass `locale` around would really simplify things.

src/handlebars/handlebarspreprocessor.js
cea2aj: Ya, I like that better

src/handlebars/handlebarspreprocessor.js
```diff
@@ -98,21 +98,31 @@ class HandlebarsPreprocessor {
...
       const count = interpolationParams.count;
-      escapedTranslatorResult = this._escapeDoubleQuotes(escapedTranslatorResult);
-      return 'ANSWERS.processTranslation(' +
-        `'${escapedTranslatorResult}', ${parsedParams}, ${count})`;
-    }
 
+      const translationPairs = Object.entries(translatorResult)
+        .reduce((params, [pluralFormIndex, pluralForm], index, array) => {
+          const escapedPluralForm = this._escapeSingleQuotes(pluralForm);
+          const accumulatedParams = params + `${pluralFormIndex}:'${escapedPluralForm}'`;
```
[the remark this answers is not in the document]
cea2aj: Ah, that's true. It is sometimes equal to 'locale'. I can make this a helper rename `pluralFormIndex`

src/handlebars/handlebarspreprocessor.js
```diff
@@ -98,21 +98,31 @@ class HandlebarsPreprocessor {
...
+          const isLastParam = (index === array.length-1);
+
+          return isLastParam ?
+            accumulatedParams :
+            accumulatedParams + ',';
+        }, '');
+
+      const translations = '{' + translationPairs + '}';
+
+      return `ANSWERS.processTranslation(${translations}, ${parsedParams}, ${count})`;
```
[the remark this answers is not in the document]
cea2aj: Ya, it just spits out something like [Object object]

src/handlebars/handlebarspreprocessor.js
cea2aj: Maybe `translationParams` or `translationPairs`?

src/handlebars/handlebarspreprocessor.js
cea2aj: I can ping Rose about this. Removing `locale` would definitely clean things up

cea2aj: How about we make locale an optional 4th parameter to the translation processor call?

tmeyer2115: Sounds good, let's see what Rose has to say. 

It is possible that the SDK is not built with the same locale as the Jambo site. For instance, we might not have a build of the SDK for a particular dialect of a language (Swiss French), just the the language itself (French). That's why we have the `sdkLocaleOverride` option. In that case, I wonder if we'd need Jambo to pass down the `locale` still to these methods. Maybe the dialect handles plurals a bit differently. If Jambo and the Theme are set up such that we still pass the dialect's locale to `ANSWERS.init`, I guess we'd be fine (provide the `init` locale is used as the default).

src/handlebars/handlebarspreprocessor.js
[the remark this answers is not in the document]
cea2aj: We decided to remove the locale from the pluralized translator result so i'll keep the name since it no longer can be equal to 'locale'

src/handlebars/handlebarspreprocessor.js
cea2aj: Spoke with Product. We decided to remove the output of the locale from jambo. For the SDK, we will use the locale that's set in the ANSWERS.init if it's not passed as an optional 4th parameter to the TranslationProcessor.process()

oshi97 (approved):
lgtm! I like that we're removing locale from translatePlural()

PR #131, opened 2020-10-07
Support for importing custom Commands and adding them to the Registry.

tmeyer2115:
This PR creates the CommandImporter class. The class imports all custom
Commands in a Jambo repo. A custom Command can be either repo or Theme
specific. Custom Commands from the repo itself are housed in a top-level
'commands' directory. The Theme's custom Commands live in the 'commands' folder
under the Theme. If the same Command is specified in both the repo and the
Theme, Jambo prefers the one from the repo. This way, Theme Commands can be
overriden.

Internally, the CommandImporter merges both 'commands' directories. It then
treats all top-level files in the merged directory as exporting a Command
for Jambo. This is a convention we will make clear in documentation. This
mechanism makes it possible for people to use 'override' to create shadows
of Theme Commands or their helpers. As soon as all the Commands are require'd,
the merged directory is deleted from the file system. In this way, it works
like 'static'. Another convention is that if custom Commands require helper
functions, those either be colocated with the Command or in the
'commands/helpers' directory.

Finally, cli.js was updated so that the imported, custom Commands are added
to the CommandRegistry. From there, the YargsFactory makes yargs commands for
them.

J=SLAP-740
TEST=manual

Tested the following:
- Created a custom Command for the Theme.
- Created a custom Command for a repo.
- Ensured that a custom Command could use helper functions or node_modules.
- If a Command is overriden in the repo, the override is preferred by Jambo.
- If a Command's helper(s) are overridden in the repo, those are preferred
  by Jambo.

### DIFF
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -27,7 +27,7 @@ class CommandImporter {
     if (commandDirectories.length > 0) {
       const mergedDirectory = this._mergeCommandDirectories(commandDirectories);
       customCommands = fs.readdirSync(mergedDirectory)
-        .map(directoryPath => path.join(process.cwd(), mergedDirectory, directoryPath))
+        .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
         .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
         .map(require);
 
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -1,0 +1,61 @@
+const path = require('path');
+const fs = require('fs-extra');
+
+/**
+ * Imports all custom {@link Command}s within a Jambo repository.
+ */
+class CommandImporter {
+  constructor(themeDir, outputDir) {
+    this._themeDir = themeDir;
+    this._outputDir = outputDir;
+  }
+
+  /**
+   * Imports custom commands from the Theme and the top-level of the Jambo
+   * repository. If a custom command is specified in both places, it is deduped, 
+   * with the override in the top-level taking priority. 
+   * 
+   * @returns {Array<Command>} The imported {@link Command}s, ready to be registered
+   *                           with Jambo.
+   */
+  import() {
+    const commandDirectories = [path.join(this._themeDir, 'commands'), 'commands']
+      .filter(fs.existsSync);
+
+    let customCommands = [];
+    if (commandDirectories.length > 0) {
+      const mergedDirectory = this._mergeCommandDirectories(commandDirectories);
+      customCommands = fs.readdirSync(mergedDirectory)
+        .map(directoryPath => path.join(process.cwd(), mergedDirectory, directoryPath))
+        .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
+        .map(require);
+
+      // Remove the merged commands directory from 'public' as it is no longer needed.
+      fs.removeSync(mergedDirectory);
+    }
+
+    return customCommands;
+  }
+
+  /**
+   * Merges the provided custom command directories together. The resulting, merged
+   * directory is in 'public'. The order in which directories are provided matters, 
+   * later ones can overwrite existing files. 
+   * 
+   * @param {Array<string>} directories The directories to merge together.
+   * @returns {string} The path of the merged output directory.
+   */
+  _mergeCommandDirectories(directories) {
+    const mergedDirectory = path.join(this._outputDir, 'commands');
+
+    // In case the merged directory has not been deleted, do so now.
+    if (fs.existsSync(mergedDirectory)) {
+      fs.removeSync(mergedDirectory);
+    }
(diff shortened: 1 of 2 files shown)


alexisgrow (dismissed):
this LGTM!

src/commands/commandimporter.js
```diff
@@ -0,0 +1,61 @@
...
+   * @returns {Array<Command>} The imported {@link Command}s, ready to be registered
+   *                           with Jambo.
+   */
+  import() {
+    const commandDirectories = [path.join(this._themeDir, 'commands'), 'commands']
+      .filter(fs.existsSync);
+
+    let customCommands = [];
+    if (commandDirectories.length > 0) {
+      const mergedDirectory = this._mergeCommandDirectories(commandDirectories);
```
oshi97, line 27: would it make sense to make an item to investigate doing the merging logic inside the javascript instead of literally copying directories? I imagine it's a little better to not make temporary folders, but probably not worth looking into right now

src/commands/commandimporter.js
```diff
@@ -0,0 +1,61 @@
...
+      customCommands = fs.readdirSync(mergedDirectory)
+        .map(directoryPath => path.join(process.cwd(), mergedDirectory, directoryPath))
```
oshi97, line 29: if we used path.resolve when creating commandDirectories and in the helper below, we would be able to avoid using process.cwd() here? I think someday we want to let people run jambo commands outside of the base directory

src/commands/commandimporter.js
tmeyer2115: I'm not sure how it would work exactly if we don't merge directories. By merging directories like this, we allow people to actually override dependencies of Theme Commands. For instance, let's say the Theme has this structure:
- commands
   - helpers
      - somehelper.js
   - customcommand.js
where customcommand.js depends on somehelper.js. By merging directories, we allow someone to override just somehelper itself. CustomCommand, assuming it's importing the helper relatively, will get the overridden implementation. I don't know how we'd allow for this without merging directories. It seems like we'd need to somehow keep track of which files satisfy which dependencies.

oshi97: I was thinking we could keep a map of filenames to filepaths and that would basically be the same thing but more code

oshi97: instead of relying on overwriting files to have that behavior

oshi97: but I'm not 100% sure it would be faster, and what we have right now is simpler

src/commands/commandimporter.js
tmeyer2115: Updated this reference here so we're not explicitly referring to `process.cwd()`. I don't think we'll get much by updating the other usages. They're relative to the top-level. This will cause issues if someone tries to run Jambo from somewhere else, but we'd have the same issue I think if we wrapped those usages in `path.resolve`. The `path.resolve` would prepend the relative paths with `cwd`, which would be invalid if someone was running Jambo not at the top-level.

src/commands/commandimporter.js
tmeyer2115: With that method, would we need to rewrite `require` statements in the Commands? Let's say I have a Command from the Theme, but I want to override some helper file it uses. I.e. my Command looks like:

```
const helper = require('./helpers/myhelper');
class CustomCommand { ...}
```

But, I have an override of `myhelper` in the repo that I want used with `CustomCommand` instead. That lives in `commands/helpers/myhelper.js`. Would I need to rewrite the require above to point to this shadowed file?

cea2aj (approved):
LGTM!

src/commands/commandimporter.js
oshi97: I see sorry I misread your earlier comment. If we support multiple command directories, would we run into issues with two command directories who import two different files with the same relative path? So a commands1/myhelper and commands2/myhelper. If we just allow for one command directory this wouldn't be an issue. Do we also want to let people require files outside of the commands directory? (I imagine restricting imports to just things in the command directory and node_modules is reasonable?)

tmeyer2115: Our idea is that there will be two `commands` directories: one in the Theme and one in the top-level of the repo. Both are optional. If you had duplicate relative file paths, the one in the top-level of the repo would win. It's similar to what would happen if someone added files to their top-level `static` directory that conflicted with the Theme. For both `static` and `commands`, if you create a file with the same relative path as something in the Theme, it becomes an override. That's a pattern the HHs should already be familiar with. If you don't want the file to be an override, you can name it differently. I think we need to make this very clear (if it's not already) in documentation. We wouldn't want people to be surprised by this behavior. 

I think we will impose the restriction that your commands can only import from within the command directory or node_modules. That feels reasonable to me too. Without that restriction, this approach won't work.

oshi97 (approved):
lgtm!

PR #133, opened 2020-10-08
Add support for incremental Typescript migration.

tmeyer2115:
This PR introduces Typescript to Jambo and makes it possible for us to do an
incremental migration. We begin the migration by converting the entry point,
cli.js, to TS. The 'allowJS' option in our tsconfig makes it so that TS will
still compile our vanilla JS. Once we've fully migrated, we can get rid of this
option.

Typescript is compiled out into regular JS. In Jambo, there is a new 'lib'
directory containing the compiled app. People who get Jambo from npm should
depend only on 'lib/cli.js'. They don't need our 'src'. I added a 'prepublish'
script and an '.npmignore' to ensure our publishes work as expected.

As an example of how to define and use types in TS, I converted the Command
class to an actual TS interface.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+lib/
 public/
 node_modules/
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+src/
+tests/
--- a/package.json
+++ b/package.json
@@ -2,8 +2,10 @@
   "name": "jambo",
   "version": "1.6.2",
   "description": "A JAMStack implementation using Handlebars",
-  "main": "index.js",
+  "main": "lib/cli.js",
   "scripts": {
+    "build": "tsc -p .",
+    "prepublish": "npm run build",
     "test": "eslint . && jest"
   },
   "repository": {
@@ -11,9 +13,9 @@
     "url": "git+https://github.com/yext/jambo.git"
   },
   "bin": {
-    "jambo": "src/cli.js"
+    "jambo": "lib/cli.js"
   },
-  "author": "nbramblett@yext.com",
+  "author": "slapshot@yext.com",
   "license": "ISC",
   "bugs": {
     "url": "https://github.com/yext/jambo/issues"
@@ -34,8 +36,10 @@
     "yargs": "^15.1.0"
   },
   "devDependencies": {
+    "@types/node": "^14.11.5",
     "eslint": "^7.8.1",
-    "jest": "^25.4.0"
+    "jest": "^25.4.0",
+    "typescript": "^4.0.3"
   },
   "jest": {
     "verbose": true,
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,8 +1,11 @@
 #!/usr/bin/env node
 
-const fs = require('file-system');
-const path = require('path');
+import fs from 'fs';
+import path from 'path';
 
+import Command from './models/commands/command';
+
+// As we cut things over to TS, we will gradually replace these with ES6 imports.
 const { parseJamboConfig } = require('./utils/jamboconfigutils');
 const { exitWithError } = require('./utils/errorutils');
 const CommandRegistry = require('./commands/commandregistry');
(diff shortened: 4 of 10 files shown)


package.json
```diff
@@ -2,18 +2,20 @@
...
   },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/yext/jambo.git"
   },
   "bin": {
-    "jambo": "src/cli.js"
+    "jambo": "lib/cli.js"
   },
-  "author": "nbramblett@yext.com",
```
oshi97, line 16: rip nik

src/commands/card/cardcreator.js
```diff
@@ -61,7 +61,7 @@ exports.CardCreator = class {
   _getRenamedCardComponent(content, customCardName) {
     const cardNameSuffix = 'CardComponent';
     const registerComponentTypeRegex = /\([\w_]+CardComponent\)/g;
-    const regexArray = [ ...content.matchAll(/componentName\s*=\s*'(.*)'/g) ];
+    const regexArray = Array.from(content.matchAll(/componentName\s*=\s*'(.*)'/g));
```
oshi97, line 64: you mean you *don't* want to overuse the spread operator? or is this a typescript thing

package.json
```diff
@@ -2,18 +2,20 @@
   "name": "jambo",
   "version": "1.6.2",
   "description": "A JAMStack implementation using Handlebars",
-  "main": "index.js",
+  "main": "lib/cli.js",
   "scripts": {
+    "build": "tsc -p .",
+    "prepublish": "npm run build",
```
oshi97, line 8: do we want some kind of dev command for watching file changes? or is that not needed (have literally 0 experience with typescript)

src/commands/card/cardcreator.js
tmeyer2115: This is a typescript thing. Does not play well with the spread operator and iterables.

package.json
tmeyer2115: I can look into adding something like that since there's actual compilation that needs to be done now

.npmignore
```diff
@@ -0,0 +1,2 @@
+src/
```
oshi97, line 1: I was wondering whether we should have this, since I noticed before when I look inside node_modules I can sometimes see both the typescript and the built javascript, and stumbled across this https://stackoverflow.com/questions/24942161/does-npm-ignore-files-listed-in-gitignore

.gitignore
```diff
@@ -1,2 +1,3 @@
+lib/
 public/
```
oshi97, line 2: kind of a separate thing but do we still need this public/ line in the gitignore?

tmeyer2115: Probably not, that's not something that exists in Jambo itself, just Jambo repos.

.npmignore
tmeyer2115: Yeah, I had to do something like this for the RTF conversion library. In the GH repo for Jambo, we don't care about `lib` since that's auto generated. I feel like people pulling Jambo down from npm probably don't care about `src`. If they did, they'd clone it from GH (which is public).

oshi97: Oh I was mainly worried about missing things from the gitignore, but I guess that would only be node_modules which apparently npm is smart enough to ignore automatically (nice!)

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
oshi97, line 1: nit but spacing is slightly off (can probably just hit the vscode auto format button)

src/commands/card/cardcreator.js
oshi97: ah that's a real bummer. my brain trembles

package.json
cea2aj: tsc has a --watch flag which we could use in a dev script

src/cli.ts
```diff
@@ -1,8 +1,11 @@
 #!/usr/bin/env node
 
-const fs = require('file-system');
-const path = require('path');
+import fs from 'fs';
```
alexisgrow, line 3: I'm thinking the answer is yes, but confirming 'fs' has the same interface as 'file-system'?

PR #134, opened 2020-10-12
jambo: remove custom jambo commands

nanhu95:
This PR removes the card and directanswercard jambo commands
from the src/commands folder. These commands will be moved
into the hitchhikers theme and be added as custom jambo commands
there.

J=SLAP-511
TEST=manual
Created a local jambo repo with card and directanswercard commands
already moved into the hitchhikers theme. Created two pages, one
card, and one directanswercard using jambo commands and used one
kind of card per page. Did a jambo build and served up the site to
see that it behaves as expected. No errors were encountered when
running jambo commands.


oshi97 (commented):
+1 from me! @tmeyer2115 could you take a quick look?

src/cli.js
```diff
@@ -26,7 +26,7 @@ if (jamboConfig && jamboConfig.dirs && jamboConfig.dirs.output) {
     new CommandImporter(jamboConfig.dirs.output);
   
   commandImporter.import().forEach(customCommand => {
-    commandRegistry.addCommand(customCommand)
+    commandRegistry.addCommand(customCommand(jamboConfig))
```
tmeyer2115, line 29: What about having the `CommandImporter` actually instantiate the `Command`? That falls in line with it's original responsibility: providing a list of custom `Command`s. Currently, it's responsible for providing a list of generating functions for the custom `Command`s, which seems harder to conceptualize.

nanhu95: Will do!

oshi97 (approved):
lgtm!

PR #135, opened 2020-10-12
Add jambo describe command

oshi97:
This commit adds the jambo describe command. When the card + da card
commands are split from jambo, those portions of the describe command
will also need to be ported.

Originally I wanted to have jambo describe pull the schema for each
command from the commands themselves, but it didn't look like there
was a way to do this without refactoring all of the built-in commands,
which seemed like a weird thing to do since there's another item in progress
to cut over and rewrite the card + da card commands.

TEST=manual,auto
J=SLAP-549

run jambo describe on a jambo site using the hh-theme, see cards,
da cards, pages, and overridable files correctly printed out

### DIFF
--- a/src/commands/describe/commanddescriber.js
+++ b/src/commands/describe/commanddescriber.js
@@ -6,7 +6,8 @@ const DescribeCommandRepoReader = require('./describecommandreporeader');
  * and their possible arguments.
  */
 class CommandDescriber {
-  constructor(jamboConfig = {}) {
+  constructor(jamboConfig = {}, commandRegistry) {
+    this.commandRegistry = commandRegistry;
     /**
      * @type {DescribeCommandRepoReader}
      */
@@ -14,14 +15,24 @@ class CommandDescriber {
   }
 
   describe() {
-    const descriptions = this._getCommandDescriptions();
+    const builtInDescriptions = this._getBuiltInDescriptions();
+    const customDescriptions = this._getCustomDescriptions();
+    const descriptions = Object.assign({}, builtInDescriptions, customDescriptions);
     console.dir(descriptions, {
       depth: null,
       maxArrayLength: null
     });
   }
 
-  _getCommandDescriptions() {
+  _getCustomDescriptions() {
+    const customDescriptions = {};
+    for (const command of this.commandRegistry.getCommands()) {
+      customDescriptions[command.getAlias()] = command.describe();
+    }
+    return customDescriptions;
+  }
+
+  _getBuiltInDescriptions() {
     const themes = this.repoReader.getThemes();
     const pageTemplates = this.repoReader.getPageTemplates();
     const themeFiles = this.repoReader.getThemeFiles();
--- a/src/commands/describe/commanddescriber.js
+++ b/src/commands/describe/commanddescriber.js
@@ -33,15 +33,15 @@ class CommandDescriber {
   }
 
   _getBuiltInDescriptions() {
-    const themes = this.repoReader.getThemes();
+    const importableThemes = this.repoReader.getImportableThemes();
     const pageTemplates = this.repoReader.getPageTemplates();
     const themeFiles = this.repoReader.getThemeFiles();
     const cards = this.repoReader.getCards();
     const daCards = this.repoReader.getDirectAnswerCards();
     return {
-      init: Descriptions.init(themes),
+      init: Descriptions.init(importableThemes),
       page: Descriptions.page(pageTemplates),
-      import: Descriptions.import(themes),
+      import: Descriptions.import(importableThemes),
       override: Descriptions.override(themeFiles),
       upgrade: Descriptions.upgrade(),
       build: Descriptions.build(),
--- a/src/commands/describe/commanddescriber.js
+++ b/src/commands/describe/commanddescriber.js
@@ -1,16 +1,16 @@
 const Descriptions = require('./descriptions');
-const RepoAnalyzer = require('./repoanalyzer');
+const DescribeCommandRepoReader = require('./describecommandreporeader');
 
 /**
  * CommandDescriber outputs JSON that describes all registered Jambo commands
  * and their possible arguments.
  */
 class CommandDescriber {
-  constructor(repoAnalyzer) {
+  constructor(jamboConfig = {}) {
     /**
-     * @type {RepoAnalyzer}
+     * @type {DescribeCommandRepoReader}
      */
-    this.repoAnalyzer = repoAnalyzer;
+    this.repoReader = new DescribeCommandRepoReader(jamboConfig);
   }
 
   describe() {
@@ -22,11 +22,11 @@ class CommandDescriber {
   }
 
   _getCommandDescriptions() {
-    const themes = this.repoAnalyzer.getThemes();
-    const pageTemplates = this.repoAnalyzer.getPageTemplates();
-    const themeFiles = this.repoAnalyzer.getThemeFiles();
-    const cards = this.repoAnalyzer.getCards();
-    const daCards = this.repoAnalyzer.getDirectAnswerCards();
+    const themes = this.repoReader.getThemes();
+    const pageTemplates = this.repoReader.getPageTemplates();
+    const themeFiles = this.repoReader.getThemeFiles();
+    const cards = this.repoReader.getCards();
+    const daCards = this.repoReader.getDirectAnswerCards();
     return {
       init: Descriptions.init(themes),
       page: Descriptions.page(pageTemplates),
--- a/src/commands/describe/commanddescriber.js
+++ b/src/commands/describe/commanddescriber.js
@@ -1,0 +1,43 @@
+const Descriptions = require('./descriptions');
+const RepoAnalyzer = require('./repoanalyzer');
+
+/**
+ * CommandDescriber outputs JSON that describes all registered Jambo commands
+ * and their possible arguments.
+ */
+class CommandDescriber {
+  constructor(repoAnalyzer) {
+    /**
+     * @type {RepoAnalyzer}
+     */
+    this.repoAnalyzer = repoAnalyzer;
+  }
+
+  describe() {
+    const descriptions = this._getCommandDescriptions();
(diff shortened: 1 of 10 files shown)


src/commands/describe/commanddescriber.js
```diff
@@ -0,0 +1,43 @@
+const Descriptions = require('./descriptions');
+const RepoAnalyzer = require('./repoanalyzer');
+
+/**
+ * CommandDescriber outputs JSON that describes all registered Jambo commands
+ * and their possible arguments.
+ */
+class CommandDescriber {
```
tmeyer2115, line 8: Our plan is for all existing built-in commands to implement the new `Command` interface. I think we have an item for that in the backlog. Since this is a new built-in command, can we make it implement this interface from the start? If both built-in and custom commands implement it, they can be treated the same way in the `YargsFactory`. Conceptually, too, both should be represented in the same way. I added a private method, called `_initialize` to the `CommandRegistry`. Once `CommandDescriber` implements `Command`, you'll want to update that method.

src/commands/describe/commanddescriber.js
```diff
@@ -0,0 +1,43 @@
...
+
+  describe() {
+    const descriptions = this._getCommandDescriptions();
+    console.dir(descriptions, {
+      depth: null,
+      maxArrayLength: null
+    });
+  }
+
+  _getCommandDescriptions() {
```
tmeyer2115, line 24: Does this work with custom `Commands`?

src/commands/describe/repoanalyzer.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs');
+const fileSystem = require('file-system');
+const path = require('path');
+
+module.exports = class RepoAnalyzer {
```
tmeyer2115, line 5: Add some JSDoc. I think it'd be good to come up with a name for this class that's a bit more descriptive as well. `Analyzer` is somewhat generic and this does something very specific.

src/commands/describe/repoanalyzer.js
```diff
@@ -0,0 +1,44 @@
...
+  constructor(jamboConfig) {
+    this.themesDir = jamboConfig.dirs.themes;
+    this.defaultTheme = jamboConfig.defaultTheme;
+  }
+
+  getThemes() {
+    return ['answers-hitchhiker-theme'];
```
tmeyer2115, line 12: Will this cause issues if they haven't imported a Theme yet before running `describe`?

src/commands/describe/commanddescriber.js
```diff
@@ -0,0 +1,43 @@
...
+  constructor(repoAnalyzer) {
+    /**
+     * @type {RepoAnalyzer}
+     */
+    this.repoAnalyzer = repoAnalyzer;
+  }
+
+  describe() {
+    const descriptions = this._getCommandDescriptions();
+    console.dir(descriptions, {
```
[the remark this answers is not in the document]
mancusi: Yeah this would be an issue. We only have the ability to read the stdout & stderr from CBD.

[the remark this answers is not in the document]
oshi97: I'm pretty sure console.dir prints to stdout, but is there some way I can check this?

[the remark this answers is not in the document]
mancusi: Hm. From the node docs this may actually be okay. If it's an issue I'll let you all know.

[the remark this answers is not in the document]
mancusi: Okay from the docs this may be fine then. If this is an issue I'll let you know.

src/commands/describe/commanddescriber.js
oshi97: It feels a little weird to have 2 places to add commands, would it make sense to register the built-in commands in YargsFactory.createCli? When I tried to the describe command to _initialize(), I found myself having to copy over YargsFactory._createCommandModule()

src/commands/describe/descriptions.js
```diff
@@ -0,0 +1,146 @@
+/**
```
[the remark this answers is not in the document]
oshi97: added a todo! there was no item to cut over the current built-in commands, so I made one

src/commands/describe/commanddescriber.js
oshi97: updated! I was thinking of doing that in a separate pr but it's not much more code to put it here

src/commands/describe/repoanalyzer.js
oshi97: added some guarding, it's not the most elegant way, but it's what the add page command does as well

src/commands/describe/describecommand.js
```diff
@@ -0,0 +1,64 @@
...
+ * DescribeCommand outputs JSON that describes all registered Jambo commands
+ * and their possible arguments.
+ */
+module.exports = class DescribeCommand {
+  constructor(jamboConfig = {}, commandRegistry) {
+    this.commandRegistry = commandRegistry;
+    /**
+     * @type {DescribeCommandRepoReader}
+     */
+    this.repoReader = new DescribeCommandRepoReader(jamboConfig);
```
tmeyer2115, line 14: If this is the only place we use `jamboConfig`, I'd suggest passing in the `DescribeCommandRepoReader` instead. That's really what this class needs, not `jamboConfig`. By passing in `DescribeCommandRepoReader`, instead of constructing it, we're more closely following dependency injection.

src/commands/describe/describecommand.js
```diff
@@ -0,0 +1,64 @@
...
+    const builtInDescriptions = this._getBuiltInDescriptions();
+    const customDescriptions = this._getCustomDescriptions();
+    const descriptions = Object.assign({}, builtInDescriptions, customDescriptions);
+    console.dir(descriptions, {
+      depth: null,
+      maxArrayLength: null
+    });
+  }
+
+  _getCustomDescriptions() {
```
tmeyer2115, line 39: JSDoc

src/commands/describe/describecommand.js
```diff
@@ -0,0 +1,64 @@
...
+    const customDescriptions = {};
+    for (const command of this.commandRegistry.getCommands()) {
+      customDescriptions[command.getAlias()] = command.describe();
+    }
+    return customDescriptions;
+  }
+
+  _getBuiltInDescriptions() {
```
tmeyer2115, line 47: JSDoc

src/commands/describe/describecommandreporeader.js
```diff
@@ -0,0 +1,73 @@
...
+const fileSystem = require('file-system');
+const path = require('path');
+
+/**
+ * DescribeCommandRepoReader holds the filesystem logic used in the describe command.
+ * TODO(SLAP-766): This class is temporary and will be removed when the built-in commands
+ * are updated to use the Command interface.
+ */
+module.exports = class DescribeCommandRepoReader {
+  constructor(jamboConfig) {
```
tmeyer2115, line 11: Does the `describe` command work before `init` is run? I could see a use case for that. Someone is maybe new to Jambo, they don't know what the commands are, what they do, etc. So, before they run `init`, they `describe`.

src/yargsfactory.js
```diff
@@ -31,6 +32,8 @@ class YargsFactory {
     this._commandRegistry.getCommands().forEach(command => {
       cli.command(this._createCommandModule(command));
     });
+    cli.command(this._createCommandModule(
+      new DescribeCommand(jamboConfig, this._commandRegistry)));
```
tmeyer2115, line 36: I think we should add this to the `CommandRegistry#_initialize`. Conceptually, this class should have one responsibility: create a Yargs instance from all the supported `Command`s. It should not have the responsibility of constructing the `Command`s as well. Additionally, it should not have to treat built-in commands differently than custom ones.

That's not how `YargsFactory` is today, unfortunately. The `_addBuiltInCommands` method is a temporary measure I added. It's an intermediary step while we cut built-in commands over to the `Command` interface. As we cut them over, we should add them to the registry and remove them from `_addBuiltInCommands`. The Registry is responsible for maintaining all supported commands. 

It is a bit weird to have the built-in commands created in two different places, but that's temporary. Once the cut-over is complete, all built-in commands will be instantiated in the registry.

src/commands/describe/describecommandreporeader.js
oshi97: updated! will work now

src/yargsfactory.js
oshi97: updated!

src/commands/describe/describecommand.js
```diff
@@ -0,0 +1,84 @@
...
+      depth: null,
+      maxArrayLength: null
+    });
+  }
+
+  /**
+   * Returns the descriptions for all commands registered
+   * with the {@link Command} interface.
+   */
+  _getCommandModuleDescriptions() {
```
tmeyer2115, line 53: I think we should rename this. The name suggests to me that we're getting descriptions of command modules, instead of descriptions of `Command`s.

src/commands/describe/describecommand.js
```diff
@@ -0,0 +1,84 @@
...
+    const descriptions = { ...builtInDescriptions, ...commandModuleDescriptions};
+    console.dir(descriptions, {
+      depth: null,
+      maxArrayLength: null
+    });
+  }
+
+  /**
+   * Returns the descriptions for all commands registered
+   * with the {@link Command} interface.
```
tmeyer2115, line 51: How about "Returns the descriptions of all registered Commands"? We're not registering commands with the interface per se.

oshi97: ah yeah that's way better

src/commands/describe/describecommand.js
oshi97: It was a little hard to name this because if I saw this registers all commands, well, aren't the built in commands also commands? I guess soon all built-in commands will also get their describe from there, though, so maybe I shouldn't fuss over something so temporary.

tmeyer2115 (approved):

PR #137, opened 2020-10-15
Fix jambo describe for import

oshi97:
type for the import --theme param should be 'singleoption' not 'string'

TEST=manual

check that jambo describe for import now has type singleoption for --theme
also double check the other commands and their params


nanhu95 (approved):
LGTM!

PR #139, opened 2020-10-19
Add extract-translations command

alexisgrow:
Convert existing extract-translations command into a class that implements Jambo's new 'Command' interface.

TEST=manual

Run `jambo --help`, see `jambo extract-translations  extract translated strings from .hbs and .js files` in the list of commands.
Run `jambo extract-translations --help` and see the following printed:
```
jambo extract-translations

extract translated strings from .hbs and .js files

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --output   the output path to extract the .pot file to
                                              [string] [default: "messages.pot"]
```
Run `jambo describe` and see the following in the list of commands described:
```
'extract-translations': {
    displayName: 'Extract Translations',
    params: {
      output: {
        displayName: 'Output Path',
        type: 'string',
        required: false,
        default: 'messages.pot'
      }
    }
  }
```
Run `jambo extract-translations` and see `messages.pot` file generated.

### DIFF
--- a/src/commands/extract-translations/jambotranslationextractor.js
+++ b/src/commands/extract-translations/jambotranslationextractor.js
@@ -1,4 +1,5 @@
 const TranslationExtractor = require('../../i18n/extractor/translationextractor');
+const { ArgumentMetadata, ArgumentType } = require('../../models/commands/argumentmetadata');
 const fsExtra = require('fs-extra');
 const fs = require('fs');
 
@@ -11,11 +12,41 @@ class JamboTranslationExtractor {
     this.extractor = new TranslationExtractor();
   }
 
+  getAlias() {
+    return 'extract-translations';
+  }
+
+  getShortDescription() {
+    return 'extract translated strings from .hbs and .js files';
+  }
+
+  args() {
+    return {
+      output: new ArgumentMetadata(
+        ArgumentType.STRING,
+        'the output path to extract the .pot file to',
+        false,
+        'messages.pot'
+      )
+    };
+  }
+
+  describe() {
+    return {
+      displayName: 'Extract Translations',
+      params: this.args()
+    };
+  }
+
+  execute(args) {
+    this._extract(args.output);
+  }
+
   /**
    * Extracts i18n strings from a jambo repo to a designed output file.
    * @param {string} outputPath
    */
-  async extract(outputPath) {
+  async _extract(outputPath) {
     const { files, directories } = this._getFilesAndDirsFromJamboConfig();
     const gitignorePaths = this._parseGitignorePaths();
     console.log(`Extracting translations to ${outputPath}`);
--- a/src/commands/extract-translations/jambotranslationextractor.js
+++ b/src/commands/extract-translations/jambotranslationextractor.js
@@ -37,7 +37,7 @@ class JamboTranslationExtractor {
     return {
       displayName: 'Extract Translations',
       params: Object.keys(args).reduce((params, alias) => {
-        params[alias] = args[alias].serialize();
+        params[alias] = args[alias].toDescribeFormat();
         return params;
       }, {})
     };
--- a/src/commands/extract-translations/jambotranslationextractor.js
+++ b/src/commands/extract-translations/jambotranslationextractor.js
@@ -22,19 +22,24 @@ class JamboTranslationExtractor {
 
   args() {
     return {
-      output: new ArgumentMetadata(
-        ArgumentType.STRING,
-        'the output path to extract the .pot file to',
-        false,
-        'messages.pot'
-      )
+      output: new ArgumentMetadata({
+        displayName: 'Output Path',
+        type: ArgumentType.STRING,
+        description: 'the output path to extract the .pot file to',
+        isRequired: false,
+        defaultValue: 'messages.pot'
+      })
     };
   }
 
   describe() {
+    const args = this.args();
     return {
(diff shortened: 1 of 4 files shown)


src/commands/extract-translations/jambotranslationextractor.js
```diff
@@ -11,11 +12,41 @@ class JamboTranslationExtractor {
...
+        false,
+        'messages.pot'
+      )
+    };
+  }
+
+  describe() {
+    return {
+      displayName: 'Extract Translations',
+      params: this.args()
```
tmeyer2115, line 37: @oshi97, what's the format expected for `params` here? I'm betting Sumo needs this data in a format other than `ArgumentMetadata`.

tmeyer2115: @alexisgrow, here's the doc outlining the `describe` command: https://docs.google.com/document/d/1psEgDwHuebRiGBvqESj3RrCGA7ECdpsPHJTGNABinwU/edit#heading=h.kgkl57hbxsec. The output of this command is JSON that will be consumed by Sumo. They're expecting things in a particular format. There's an example of the output on page 3 of the doc. I don't think they will be able to parse serialized `ArgumentMetadata`s directly.

alexisgrow: Oh that makes sense. What do you think about us adding a serialization method to the ArgumentMetadata class that outputs metadata in the format that works for Sumo? It seems like a potential place for typos if we have to manually specify the metadata in both formats in the command implementation

oshi97: I think after we cut over all the built-in commands, it would be great to have the params part of describe generated from the args (kind of like this)! but right now we don't have anything to do that

src/commands/extract-translations/jambotranslationextractor.js
```diff
@@ -11,11 +12,41 @@ class JamboTranslationExtractor {
...
+      output: new ArgumentMetadata(
+        ArgumentType.STRING,
+        'the output path to extract the .pot file to',
+        false,
+        'messages.pot'
+      )
+    };
+  }
+
+  describe() {
```
oshi97, line 34: after updating the describe params, could we add this to the describe unit test?

src/commands/extract-translations/jambotranslationextractor.js
tmeyer2115: I like the idea of adding a serialization method to `ArgumentMetadata`! You're right, it's better to have that logic centralized, rather than spread all over. If you want to skip this and do it manually for now in `JamboTranslationExtractor`, no worries. We'd just make an item.

src/models/commands/argumentmetadata.js
```diff
@@ -42,11 +43,29 @@ class ArgumentMetadata {
...
 
+  /**
+   * @returns {string} The display name for the argument.
+   */
+  getDisplayName() {
+    return this._displayName;
+  }
+
+  /**
+   * @returns {Object} The serialized ArgumentMetadata
```
tmeyer2115, line 60: Could this be JSON? Serialization suggest a stringified representation of something. If serializing to a string won't work here, maybe we could call the method something else? We'd want a name that indicates this is for use with the `describe` command.

alexisgrow: Makes sense! As you mentioned JSON serialization won't work here (it would look like this, which is not what we want:
```
  'extract-translations': {
    displayName: 'Extract Translations',
    params: {
      output: '{"displayName":"Output Path","type":"string","required":false,"default":"messages.pot"}'
    }
  }
```
because then only a portion of the describe output object would be serialized) so I renamed this to `toDescribeFormat()`. Let me know what you think!

oshi97 (commented):
+1! bump @tmeyer2115

tests/commands/describe/describecommand.js
```diff
@@ -81,10 +81,29 @@ const mockDirectAnswerCardCommand = {
...
 
+const mockExtractTranslationsCommand = {
+  getAlias() {
+    return 'extract-translations';
+  },
+  describe() {
+    return {
+      displayName: 'Extract Translations',
+      params: {
+        output: { displayName: 'Output Path' }
```
oshi97, line 92: oh can we add type: 'string' here as well? we don't have any idea of a default type (maybe it should be 'string'? we can ask sumo)

oshi97: oh could we also include a test-case for extract-translations? just a really dumb pass through like the rest of these

alexisgrow: We've got it in the non-mock version of the extract translations command, I can add it here! But I was going back and forth because I don't think copying over all of the params from the implementation to this mock implementation provides any more coverage, is that right? I could use the real implementation rather than a mock?

tmeyer2115 (dismissed):

tests/commands/describe/describecommand.js
oshi97: ah yeah we probably don't need a mock here, probably better to just get an actual instance of the command. okay to skip this though!

alexisgrow: Sweet I'll update, adding test coverage for this totally makes sense to me as part of this PR! ✨

oshi97 (approved):

PR #140, opened 2020-10-19
Default relativePath to "." instead of ""

oshi97:
This commit updates relativePath to be "." instead of ""
when the page output path is the root output dir.

J=SLAP-773
TEST=manual

check that on a new hh-theme page with no locale_config,
bundle.js and a favicon can load correctly

### DIFF
--- a/src/commands/build/pagewriter.js
+++ b/src/commands/build/pagewriter.js
@@ -94,6 +94,6 @@ module.exports = class PageWriter {
    * @returns {String}
    */
   _calculateRelativePath(filePath) {
-    return path.relative(path.dirname(filePath), '');
+    return path.relative(path.dirname(filePath), '') || '.';
   }
 }


src/commands/build/pagewriter.js
```diff
@@ -94,6 +94,6 @@ module.exports = class PageWriter {
    * @returns {String}
    */
   _calculateRelativePath(filePath) {
-    return path.relative(path.dirname(filePath), '');
+    return path.relative(path.dirname(filePath), '') || '.';
```
tmeyer2115, line 97: Should this be `path.relative(path.dirname(filePath), '.')` or does that matter?

oshi97: naw either is fine (tried the above too originally) , both just point to the root directory
https://nodejs.org/api/path.html#path_path_relative_from_to

tmeyer2115 (approved):

PR #141, opened 2020-10-19
Swap order of params in CommandImporter constructor

alexisgrow:
For cases when commands are run pre-theme import, the CommandImporter would fail (an example of this is the `jambo import --theme` command). This is because the second case in [this ternary](https://github.com/yext/jambo/blob/master/src/cli.js#L27) doesn't have a theme to pass into the constructor. The theme param should be last since it's optional.

TEST=manual

Run `jambo import --theme answers-hitchhiker-theme` before this, see error. Run again after, see theme import successfully.


oshi97 (approved):
lgtm!

PR #142, opened 2020-10-19
I18n pre rebase copy

alexisgrow:


alexisgrow:
Was just using this to diff the rebased version and a version with master merged into it. All LG, so closing.

PR #144, opened 2020-10-19
Canonicalize locales

alexisgrow:
Adds canonicalization logic to normalize casing and locale code separators. This logic converts the language in the locale to lowercase and the region (if provided) to uppercase. Replaces dashes with underscores.

TEST=manual
J=SLAP-705

Test with varied casing in config names and page names.

### DIFF
--- a/src/models/localizationconfig.js
+++ b/src/models/localizationconfig.js
@@ -1,4 +1,5 @@
 const { NO_LOCALE } = require('../constants');
+const { canonicalizeLocale } = require('../utils/i18nutils');
 const UserError = require('../errors/usererror');
 
 /**
@@ -25,7 +26,11 @@ module.exports = class LocalizationConfig {
      *   ...
      * }
      */
-    this._localeToConfig = config.localeConfig || {};
+    this._localeToConfig = {};
+    for (const [localeCode, localeConfig] of Object.entries(config.localeConfig || {})) {
+      const normalizedLocale = canonicalizeLocale(localeCode);
+      this._localeToConfig[normalizedLocale] = localeConfig;
+    };
 
     if (this.hasConfig() && !this._localeToConfig[this._defaultLocale]) {
       throw new UserError(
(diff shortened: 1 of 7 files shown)


cea2aj:
This looks good to me. I think we're also going to want a corresponding PR in the SDK to lowercase the locale bundles: https://github.com/yext/answers/blob/master/conf/i18n/constants.js

tmeyer2115:
> This looks good to me. I think we're also going to want a corresponding PR in the SDK to lowercase the locale bundles: https://github.com/yext/answers/blob/master/conf/i18n/constants.js

We want to avoid any SDK changes because that means cutting a new release. Is there a clean way we can accomplish this without updating the SDK? What if we have the canonical form be upper case (i.e. en_GB)? The HH can still supply it in whatever way they want, but we canonicalize it.

alexisgrow:
> > This looks good to me. I think we're also going to want a corresponding PR in the SDK to lowercase the locale bundles: https://github.com/yext/answers/blob/master/conf/i18n/constants.js
> 
> We want to avoid any SDK changes because that means cutting a new release. Is there a clean way we can accomplish this without updating the SDK? What if we have the canonical form be upper case (i.e. en_GB)? The HH can still supply it in whatever way they want, but we canonicalize it.

Yeah we can do that! Updated

src/models/localizationconfig.js
```diff
@@ -79,9 +84,8 @@ module.exports = class LocalizationConfig {
    * @returns {function}
    */
   getUrlFormatter(locale) {
-    // TODO (agrow) this assumes language and region are separated by a "-" (e.g. en-US)
     const language = locale
-      ? locale.substring(0, locale.lastIndexOf('-')) || locale
+      ? locale.substring(0, locale.indexOf('-')) || locale
```
oshi97, line 88: would it be simpler to do locale.split('-')[0]? probably just a personal taste thing

oshi97 (approved):
this lgtm!

src/models/localizationconfig.js
alexisgrow: That makes sense to me! I'll update!

cea2aj:
The locale bundles we are hosting right now are using underscores between the language and locale. Can we use that for now to stay consistent?

cea2aj (approved):
LG!

PR #145, opened 2020-10-20
Translation loading improvements

cea2aj:
Modify translation file loading so that all custom translation files are not strictly necessary

Modify localFileParser so that it no longer tries to default the file name if just a locale is provided. Modify the site generator to always load theme translations, and to only load custom translations if the "translations" dir is configured. A custom translation file for each configured locale is no longer required.

J=SLAP-706
TEST=manual

Test a jambo build with and without custom translation dirs and custom translations file paths. Confirm translations are correct. Test that only some of the translation files for the configured locales can be present when building a site.

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -389,10 +389,7 @@ exports.SitesGenerator = class {
     const translations = {};
 
     for (const locale of locales) {
-      const translationFileOverride = localizationConfig.getTranslationFile(locale);
-      const translationFileName = translationFileOverride
-        ? translationFileOverride
-        : `${locale}.po`
+      const translationFileName = localizationConfig.getTranslationFile(locale) || `${locale}.po`;
       const translationFilePath = path.join(translationsDir, translationFileName);
       const isDefaultLocale = (locale === localizationConfig.getDefaultLocale());
       if (!isDefaultLocale && fs.existsSync(translationFilePath)) {
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -379,18 +379,25 @@ exports.SitesGenerator = class {
    * @returns {Object<string, Object>} A map of locale to formatted translations.
    */
   async _extractCustomTranslations(locales, localizationConfig) {
-    const translationsDir = this.config.dirs.translations
-      ? this.config.dirs.translations
-      : 'translations';
+    const translationsDir = this.config.dirs.translations;
+
+    if (!translationsDir) {
+      return {};
+    }
 
     const localFileParser = new LocalFileParser(translationsDir);
     const translations = {};
 
     for (const locale of locales) {
+      const translationFileOverride = localizationConfig.getTranslationFile(locale);
+      const translationFileName = translationFileOverride
+        ? translationFileOverride
+        : `${locale}.po`
+      const translationFilePath = path.join(translationsDir, translationFileName);
       const isDefaultLocale = (locale === localizationConfig.getDefaultLocale());
-      if (!isDefaultLocale) {
+      if (!isDefaultLocale && fs.existsSync(translationFilePath)) {
         const localeTranslations = await localFileParser
-          .fetch(locale, localizationConfig.getTranslationFile(locale));
+          .fetch(locale, translationFileName);
         translations[locale] = { translation: localeTranslations };
       }
     }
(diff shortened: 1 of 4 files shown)


src/i18n/translationfetchers/localfileparser.js
```diff
@@ -21,8 +21,10 @@ class LocalFileParser {
   }
 
   /**
-   * Extracts a locale's translations from the local filesystem. If no such
-   * translations exist, an empty object is returned.
+   * Extracts a locale's translations from the local filesystem. If just locale is passed
+   * and the translation file doesn't exist, an empty object is returned. If a
+   * translationFilePath is specified and the translation files doesn't exist, the
+   * function rejects with a promise
```
tmeyer2115, line 27: I think the method is doing too much now and has become hard to describe. There are three different conditional branches. Plus, it attempts to default the `translationFilePath`. What I'm suggesting is:
- We remove the `translationFilePath` defaulting logic. It's on the caller to do that if they want it.
- The method checks that the `translationFilePath` exists. If it does, we proceed. Otherwise, an error is thrown. The caller can do with that error whatever it is they'd like. 

Defaulting the file path and then changing behavior based on if we have defaulted it complicates this method. Arguably, it makes it less general as well. There may be some future caller of the method who always wants a `UserError`, even if the file path was defaulted.

src/commands/build/sitesgenerator.js
```diff
@@ -380,11 +379,16 @@ exports.SitesGenerator = class {
    * @returns {Object<string, Object>} A map of locale to formatted translations.
    */
   async _extractCustomTranslations(locales, localizationConfig) {
-    const localFileParser = new LocalFileParser(this.config.dirs.translations);
+    const translationsDir = this.config.dirs.translations
+      ? this.config.dirs.translations
+      : 'translations';
```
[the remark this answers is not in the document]
cea2aj: got it!

src/i18n/translationfetchers/localfileparser.js
cea2aj: For the first bullet point, are you suggesting that we remove the 'locale' param and make the only parameter of the fetch be 'translationFilePath'?

tmeyer2115: We can't get rid of the `locale` param as we still need it in `fetch`. But, we should not try and default the file name using it. Arguably, I shouldn't have had his method do that to begin with. It gives this method too much responsibility, especially considering the complex validation that happens when we do default the path.

src/commands/build/sitesgenerator.js
```diff
@@ -380,13 +379,25 @@ exports.SitesGenerator = class {
...
+      return {};
+    }
+
+    const localFileParser = new LocalFileParser(translationsDir);
     const translations = {};
 
     for (const locale of locales) {
-      if (locale !== localizationConfig.getDefaultLocale()) {
+      const translationFileOverride = localizationConfig.getTranslationFile(locale);
+      const translationFileName = translationFileOverride
```
tmeyer2115, line 393: Small nit: can we simplify this? Something like:

```
const translationFileName = localizationConfig.getTranslationFile(locale) || `${locale}.po`;
```

src/commands/build/sitesgenerator.js
```diff
@@ -380,13 +379,25 @@ exports.SitesGenerator = class {
...
+        ? translationFileOverride
+        : `${locale}.po`
+      const translationFilePath = path.join(translationsDir, translationFileName);
+      const isDefaultLocale = (locale === localizationConfig.getDefaultLocale());
+      if (!isDefaultLocale && fs.existsSync(translationFilePath)) {
         const localeTranslations = await localFileParser
-          .fetch(locale, localizationConfig.getTranslationFile(locale));
+          .fetch(locale, translationFileName);
         translations[locale] = { translation: localeTranslations };
```
tmeyer2115, line 401: Do we need to set `translations[locale] = {}` for those languages/locales that don't have custom translations? Or are we fine to just leave those out of `translations` all together?

src/commands/build/sitesgenerator.js
cea2aj: sure!

src/commands/build/sitesgenerator.js
cea2aj: Nope, it's not necessary. These get merged with the translations in the theme, so `translations[locale]` gets set during that merge. If the theme translations also don't have a particular locale, nothing breaks and the translations simply fallback to english (I tested this).

tmeyer2115 (approved):

PR #146, opened 2020-10-20
Canonicalize the default locale

oshi97:
Was running into this issue on SGS, using a the locale_config
from the diageo repo, which had a non canonical default locale.

J=SLAP-776
TEST=manual

see that builds run again


alexisgrow (approved):
LGTM!

PR #148, opened 2020-10-20
Exit code improvements

cea2aj:
Catch and handle all exit codes at the top-level of jambo

During the localized-pages rebase, some promise rejections were unhandled and jambo was incorrectly exiting with a code of 0. This was happening because the previous exit-code architecture made the commands responsible for catching and handling thrown exceptions and unhandled promise rejections. This change removes this responsibility from the commands and instead handles all errors and unhandled rejections at the top level.

J=SLAP-796
TEST=manual

Test various Jambo commands and confirm they work and exit with the right codes. Confirm that jambo does not exit with a code of 0 for unhandled promise rejections.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -9,8 +9,6 @@ const CommandRegistry = require('./commands/commandregistry');
 const YargsFactory = require('./yargsfactory');
 const CommandImporter = require('./commands/commandimporter');
 
-let jamboConfig;
-
 // Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
 process.on('unhandledRejection', err => {
   exitWithError(err);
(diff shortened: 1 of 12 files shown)


src/cli.js
```diff
@@ -11,11 +11,15 @@ const CommandImporter = require('./commands/commandimporter');
...
-}
+// Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
+process.on('unhandledRejection', err => {
+  exitWithError(err);
+});
+process.on('uncaughtException', err => {
+  exitWithError(err);
+});
+
+jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
```
alexisgrow, line 22: nit: can this be a const?

alexisgrow (commented):
One comment, otherwise LGTM

src/cli.js
cea2aj: sure

nanhu95 (approved):
LGTM!

oshi97 (approved):
looks great!

alexisgrow (approved):
LGTM!

PR #150, opened 2020-10-23
commands: cut over built in commands to use command interface

nanhu95:
Cuts over all of the build in Jambo commands to using the command
interface.

Removes describecommandreporeader.js and descriptions.js since
they are no longer needed.

Updates automated tests for describecommand.js.

J=SLAP-786
TEST=auto, manual
Ran npm run test and saw tests pass.
For each built in command, ran `jambo [cmd] --help` to verify
that the correct information shows up, ran `jambo [cmd]` with
valid arguments to see that the command works as expected, and
ran `jambo describe` to see that the information returned is
correct.


src/commands/page/add/pagecommand.js
```diff
@@ -0,0 +1,88 @@
...
+        name: {
+          displayName: 'Page Name',
+          type: 'string',
+          required: true
+        },
+        template: {
+          displayName: 'Page Template',
+          type: 'singleoption',
+          options: pageTemplates
+        }
```
alexisgrow, line 57: "layout" is listed as an argument in the `args()` function, should it be here as well?

src/commands/page/add/pagecommand.js
```diff
@@ -0,0 +1,88 @@
...
+      { ...args, theme: this.defaultTheme });
+    const pageScaffolder = new PageScaffolder(this.jamboConfig);
+    try {
+      pageScaffolder.create(pageConfiguration);
+    } catch (err) {
+      throw new UserError('Failed to add page', err.stack);
+    }
+
+
+    console.log(this.jamboConfig)
```
alexisgrow, line 84: Can this be removed?

tests/commands/describe/describecommand.js
```diff
@@ -81,17 +65,111 @@ const mockDirectAnswerCardCommand = {
...
+}
 const extractTranslationsCommand = new JamboTranslationExtractor({});
+const buildCommand = new BuildCommand({});
+const upgradeCommand = new UpgradeCommand(mockJamboConfig);
 
 describe('DescribeCommand works correctly', () => {
   new DescribeCommand(
     () => [
+      buildCommand,
+      mockInitCommand,
```
alexisgrow, line 164: Do you think it would be worth using the implementations of these commands here to get different coverage?

src/commands/page/add/pagecommand.js
nanhu95: Will add!

tests/commands/describe/describecommand.js
nanhu95: My thought is that since our goal is to test the describe command and not any other command, it's okay to mock the other commands. Also now that each individual command is responsible for file system traversal, as opposed to having a repoReader handle it, I don't think it'll be all that straighforward to allow actually running that code when testing. Open to other suggestions if you feel strongly about this, though!

alexisgrow: That makes sense to me! In that case, it seems like mocking one command is enough for test cases here, is that right? For example, in the "describes init" test and the "describes page" test, the describe command is tested in the same way, the only additional coverage (I think) is that its using a different mocked test fixture

src/commands/build/buildcommand.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  describe() {
+    return {
+      displayName: 'Build Pages'
+    }
+  }
+
+  execute(args) {
+    const sitesGenerator = new SitesGenerator(this.jamboConfig);
```
tmeyer2115, line 38: For the sake of dependency injection, I think we should have `BuildCommand` take in a `SitesGenerator`. Passing the `jamboConfig` to this class is a bit roundabout as the config is only used to make a `SitesGenerator`.

src/commands/build/buildcommand.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  getShortDescription() {
+    return 'build the static pages for the site';
+  }
+
+  args() {
+    return {
+      jsonEnvVars: new ArgumentMetadata({
+        type: ArgumentType.ARRAY,
```
tmeyer2115, line 24: I think we should add an `itemType` attribute to `ArgumentMetadata` as well. It would be optional and would describe the `ArgumentType` of the array elements. I believe Yargs can surface that kind of information for array arguments when `[command] --help` is called.

tests/commands/describe/describecommand.js
nanhu95: That's a great point! I'll go ahead and reduce the number of test cases.

src/commands/override/overridecommand.js
```diff
@@ -0,0 +1,73 @@
...
+          options: themeFiles
+        }
+      }
+    }
+  }
+
+  /**
+   * @returns {Array<string>} all theme files that can be overridden
+   */
+  _getThemeFiles() {
```
tmeyer2115, line 52: The `override` command doesn't restrict people to overriding a single file at a time. They can pass a sub-folder of the Theme as the path. If they do, they will get a shadow of the folder and all its content. So, we want to list all the sub-folders in the Theme as options too. This may require some coordination with Sumo. There maybe better ways of displaying this than a list of options (potential showing the Theme filesystem via a nice UI). cc: @mancusi

src/commands/page/add/pagecommand.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * PageCommand registers a new page with the specified name to be built by Jambo.
+ */
+class PageCommand {
+  constructor(jamboConfig = {}) {
+    this.jamboConfig = jamboConfig;
+    this.themesDir = jamboConfig.dirs && jamboConfig.dirs.themes;
+    this.defaultTheme = jamboConfig.defaultTheme;
+  }
+  getAlias() {
```
tmeyer2115, line 15: Newline before `getAlias`.

src/commands/page/add/pagecommand.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  args() {
+    return {
+      name: new ArgumentMetadata({
+        type: ArgumentType.STRING,
+        description: 'name for the new files',
+        isRequired: true
+      }),
+      layout: new ArgumentMetadata({
```
tmeyer2115, line 30: Here and below, I'd suggest leaving out `layout`. It's a really old, deprecated option. I don't think it works properly in `PageScaffolder` any more. We just haven't removed it yet. Nobody should be using it.

src/commands/page/add/pagecommand.js
```diff
@@ -0,0 +1,102 @@
...
+      return [];
+    }
+    const pageLayoutDir = path.resolve(this.themesDir, this.defaultTheme, 'layouts');
+    return fs.readdirSync(pageLayoutDir);
+  }
+
+  execute(args) {
+    const pageConfiguration = new PageConfiguration(
+      { ...args, theme: this.defaultTheme });
+    const pageScaffolder = new PageScaffolder(this.jamboConfig);
```
tmeyer2115, line 93: Again, for DI, let's make PageScaffolder a constructor param as well. We'll still need the `jamboConfig` as a constructor param too, but I think it will help with testability if we inject `PageScaffolder`.

src/commands/page/add/pagecommand.js
nanhu95: Ah gotcha! Thanks for the explanation, will remove.

src/commands/build/buildcommand.js
nanhu95: Looking at the yargs doc [here](https://yargs.js.org/docs/#api-reference-optionkey-opt), it doesn't look like we have the ability to specify the type of the elements in an array. Please let me know if I've missed something!

tmeyer2115: Oh, okay! I think having an optional `itemType` could still be useful. That info may come in handy in the future (we might switch to some other CLI builder that does support it).

nanhu95: Yep that makes sense, will add!

src/commands/page/add/pageconfiguration.js
```diff
@@ -0,0 +1,26 @@
+class PageConfiguration {
```
tmeyer2115, line 1: Add some JSDoc describing this new class

src/models/commands/argumentmetadata.js
```diff
@@ -9,14 +9,29 @@ const ArgumentType = {
...
+ * 
+ * ItemType is an optional parameter to be used when an array argument
+ * is specified.
+ */
+const ItemType = {
+  STRING: 'string',
+  NUMBER: 'number',
+  BOOLEAN: 'boolean'
+}
+Object.freeze(ItemType);
```
tmeyer2115, line 24: What do you think of just reusing `ArgumentType` for the `itemType`? That was my thought initially.

nanhu95: Sounds good! I was originally thinking that it might be better to differentiate between an argument's type and the type of elements within an argument, but now that you say it, I don't think there will be confusion.

src/commands/override/overridecommand.js
nanhu95: Created this [item](https://yexttest.atlassian.net/browse/SLAP-807) to track this work.

src/commands/describe/describecommand.js
```diff
@@ -37,9 +30,8 @@ module.exports = class DescribeCommand {
   }
 
   execute() {
-    const builtInDescriptions = this._getBuiltInDescriptions();
     const commandModuleDescriptions = this._getCommandDescriptions();
```
oshi97, line 33: can we get rid of one of these duplicated variables?

oshi97 (commented):
+1 from me! will defer to alexis/tom

src/commands/override/overridecommand.js
```diff
@@ -0,0 +1,73 @@
...
+const fileSystem = require('file-system');
+const { ShadowConfiguration, ThemeShadower } = require('./themeshadower');
+const { ArgumentMetadata, ArgumentType } = require('../../models/commands/argumentmetadata');
+
+/**
+ * OverrideCommand overrides a specific file in the theme.
+ */
+class OverrideCommand {
+  constructor(jamboConfig = {}) {
+    this.jamboConfig = jamboConfig
```
alexisgrow, line 11: nit: Could this have a semi-colon at the end?

src/yargsfactory.js
```diff
@@ -34,137 +23,6 @@ class YargsFactory {
...
-            .option('template', { description: 'template to use within theme' });
-        },
-        argv => {
-          const pageConfiguration = new addPageCommand.PageConfiguration(
-            { ...argv, theme: jamboConfig.defaultTheme });
-          const pageScaffolder = new addPageCommand.PageScaffolder(jamboConfig);
-          try {
-            pageScaffolder.create(pageConfiguration);
-          } catch (err) {
-            throw new UserError('Failed to add page', err.stack);
```
alexisgrow, line 117: It looks like this had several errors that were cast as UserErrors vs SystemErrors, was that behavior preserved in the new command setup?

nanhu95: Looks I missed one in buildcommand.js. It's preserved everywhere else, though.

alexisgrow (approved):
this LGTM!

PR #151, opened 2020-10-27
commands: give upgrade command optional branch arg

nanhu95:
Makes changes to provide the `jambo upgrade` command an optional
`branch` argument. This argument will specify a branch to upgrade
the theme to, instead of always upgrading to the HEAD of master.

J=SLAP-789
TEST=manual
Performed a `jambo upgrade --branch [branch_name]` using several
different branches to verify that the theme was upgraded to have
changes from the specified branch.

### DIFF
--- a/src/commands/describe/describecommand.js
+++ b/src/commands/describe/describecommand.js
@@ -55,8 +55,6 @@ module.exports = class DescribeCommand {
         }
       }
     );
-    return Promise.all(describePromises).then(
-      () => descriptions
-    );
+    return Promise.all(describePromises).then(() => descriptions);
   }
 }
--- a/src/commands/describe/describecommand.js
+++ b/src/commands/describe/describecommand.js
@@ -29,8 +29,8 @@ module.exports = class DescribeCommand {
     return {};
   }
 
-  execute() {
-    console.dir(this._getCommandDescriptions(), {
+  async execute() {
+    console.dir(await this._getCommandDescriptions(), {
       depth: null,
       maxArrayLength: null
     });
@@ -41,11 +41,22 @@ module.exports = class DescribeCommand {
    */
   _getCommandDescriptions() {
     const descriptions = {};
-    for (const command of this.getCommands()) {
-      if (command.getAlias() !== 'describe') {
-        descriptions[command.getAlias()] = command.describe();
+    const describePromises = this.getCommands().map(
+      command => {
+        const describeValue = command.describe();
+        if (describeValue.then && typeof describeValue.then === 'function') {
+          return describeValue.then(
+            (value) => { descriptions[command.getAlias()] = value; }
+          );
+        } else {
+          if (command.getAlias() !== 'describe') {
+            descriptions[command.getAlias()] = describeValue;
+          }
+        }
       }
-    }
-    return descriptions;
+    );
+    return Promise.all(describePromises).then(
+      () => descriptions
+    );
   }
 }
--- a/src/commands/upgrade/themeupgrader.js
+++ b/src/commands/upgrade/themeupgrader.js
@@ -18,7 +18,7 @@ const git = simpleGit();
  * then handles the upgrade accordingly.
  */
 class ThemeUpgrader {
-  constructor(jamboConfig) {
+  constructor(jamboConfig = {}) {
     this.jamboConfig = jamboConfig;
     this._themesDir = jamboConfig.dirs && jamboConfig.dirs.themes;
     this.upgradeScript = 'upgrade.js';
@@ -43,6 +43,11 @@ class ThemeUpgrader {
         type: ArgumentType.BOOLEAN,
         description: 'whether to pass the --isLegacy flag to ./upgrade.js',
         isRequired: false
+      }),
+      branch: new ArgumentMetadata({
+        type: ArgumentType.STRING,
+        description: 'the branch of the theme to upgrade to',
+        isRequired: false
       })
     }
   }
@@ -58,13 +63,20 @@ class ThemeUpgrader {
         disableScript: {
           displayName: 'Disable Upgrade Script',
           type: 'boolean'
+        },
+        branch: {
+          displayName: 'Branch of theme to upgrade to',
+          type: 'string'
         }
       }
     }
   }
 
   execute(args) {
-    this._upgrade(this.jamboConfig.defaultTheme, args.disableScript, args.isLegacy)
+    this._upgrade(this.jamboConfig.defaultTheme,
+      args.disableScript,
+      args.isLegacy,
+      args.branch)
       .catch(err => {
         if (isCustomError(err)) {
           throw err;
@@ -78,16 +90,18 @@ class ThemeUpgrader {
    * @param {string} themeName
    * @param {boolean} disableScript
    * @param {booolean} isLegacy
+   * @param {string} branch
    */
-  async _upgrade(themeName, disableScript, isLegacy) {
+  async _upgrade(themeName, disableScript, isLegacy, branch) {
     const themePath = path.join(this._themesDir, themeName);
     if (!fs.existsSync(themePath)) {
       throw new UserError(
         `Theme "${themeName}" not found within the "${this._themesDir}" folder`);
     }
+    const updateBranch = branch ? branch : 'master';
     await this._isGitSubmodule(themePath)
-      ? await this._upgradeSubmodule(themePath)
-      : await this._recloneTheme(themeName, themePath);
+      ? await this._upgradeSubmodule(themePath, updateBranch)
+      : await this._recloneTheme(themeName, themePath, updateBranch);
     const upgradeScriptPath = path.join(themePath, this.upgradeScript);
     if (!disableScript) {
       this._executePostUpgradeScript(upgradeScriptPath, isLegacy);
@@ -128,21 +142,25 @@ class ThemeUpgrader {
   /**
    * Calls "git update --remote" on the given submodule path, which
    * updates the given submodule to the most recent version of the branch
-   * it is set to track (defaults to master).
+   * it is set to track (defaults to master). If a branch is specified,
+   * the given submodule will be updated to the provided branch.
    * @param {string} submodulePath
+   * @param {string} updateBranch
    */
-  async _upgradeSubmodule(submodulePath) {
-    await git.submoduleUpdate(['--remote', submodulePath])
+  async _upgradeSubmodule(submodulePath, updateBranch) {
+    await git.subModule(['set-branch', '--branch', updateBranch]);
+    await git.submoduleUpdate(['--remote', submodulePath]);
   }
 
   /**
    * @param {string} themeName
    * @param {string} themePath
+   * @param {string} updateBranch
    */
-  async _recloneTheme(themeName, themePath) {
+  async _recloneTheme(themeName, themePath, updateBranch) {
     await fs.remove(themePath);
     const themeRepoURL = getRepoForTheme(themeName);
-    await git.clone(themeRepoURL, themePath);
+    await git.clone(themeRepoURL, themePath, ['--branch', updateBranch]);
   }
 
   /**
--- a/src/commands/upgrade/themeupgrader.js
+++ b/src/commands/upgrade/themeupgrader.js
@@ -78,7 +78,10 @@ class ThemeUpgrader {
     const branchesGit = simpleGit(
       path.join(this._themesDir, this.jamboConfig.defaultTheme));
     const branches = await branchesGit.branch(['--remote']);
-    return branches.all.map(branch => branch.replace('origin/', ''))
+    return branches.all.map(branch => 
+      // regex replaces 'origin/' only at the beginning of a string
+      branch.replace(/^(origin\/)/, '')
+    );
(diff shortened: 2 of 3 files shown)


src/commands/upgrade/themeupgrader.js
```diff
@@ -58,13 +63,20 @@ class ThemeUpgrader {
         disableScript: {
           displayName: 'Disable Upgrade Script',
           type: 'boolean'
+        },
+        branch: {
+          displayName: 'Branch of theme to upgrade to',
+          type: 'string'
```
oshi97, line 69: we may want this to be type: 'singleoption', and return an array of the possible branches

src/commands/upgrade/themeupgrader.js
```diff
@@ -18,9 +18,9 @@ const git = simpleGit();
  * then handles the upgrade accordingly.
  */
 class ThemeUpgrader {
-  constructor(jamboConfig) {
+  constructor(jamboConfig = {}) {
     this.jamboConfig = jamboConfig;
-    this._themesDir = jamboConfig.dirs.themes;
+    this._themesDir = jamboConfig.dirs && jamboConfig.dirs.themes;
```
oshi97, line 23: should we throw an error if there is no themes directory in the jambo config?

src/commands/upgrade/themeupgrader.js
```diff
@@ -58,13 +63,20 @@ class ThemeUpgrader {
...
+          type: 'string'
         }
       }
     }
   }
 
   execute(args) {
-    this._upgrade(this.jamboConfig.defaultTheme, args.disableScript, args.isLegacy)
+    this._upgrade(this.jamboConfig.defaultTheme,
+      args.disableScript,
```
oshi97, line 77: nit but maybe we can just pass in an args object as the second parameter?

src/commands/upgrade/themeupgrader.js
```diff
@@ -128,21 +142,25 @@ class ThemeUpgrader {
   /**
    * Calls "git update --remote" on the given submodule path, which
    * updates the given submodule to the most recent version of the branch
-   * it is set to track (defaults to master).
+   * it is set to track (defaults to master). If a branch is specified,
```
oshi97, line 145: nit but it's a little weird that the comment for this function also describes the behavior of the function that calls it.

src/commands/upgrade/themeupgrader.js
```diff
@@ -128,21 +142,25 @@ class ThemeUpgrader {
...
+   * the given submodule will be updated to the provided branch.
    * @param {string} submodulePath
+   * @param {string} updateBranch
    */
-  async _upgradeSubmodule(submodulePath) {
-    await git.submoduleUpdate(['--remote', submodulePath])
+  async _upgradeSubmodule(submodulePath, updateBranch) {
+    await git.subModule(['set-branch', '--branch', updateBranch]);
```
oshi97, line 151: instead of defaulting to master, what if we just didn't call set-branch if no --branch arg was given?

Let's say somebody was on a branch other than master, e.g. develop, then called jambo upgrade without specifying --branch. Would their theme branch be changed from develop to master?

src/commands/upgrade/themeupgrader.js
nanhu95: We're currently not doing that for any other command that uses the themes directory. My understanding is that once jambo init has been run we're guaranteed to have a valid jambo config from `jambo.json` (please correct me if I'm wrong), so we're safe here.

src/commands/upgrade/themeupgrader.js
nanhu95: That's a good point, will make the change!

src/commands/upgrade/themeupgrader.js
nanhu95: Hmm I see what you mean, although I'm not sure how best to rephrase the comment. Because for here, we're specifically talking about upgrading a submodule, while `_upgrade` is more general and covers both submodules and not. Open to suggestions!

src/commands/upgrade/themeupgrader.js
```diff
@@ -78,16 +90,18 @@ class ThemeUpgrader {
...
+   * @param {string} branch
    */
-  async _upgrade(themeName, disableScript, isLegacy) {
+  async _upgrade(themeName, disableScript, isLegacy, branch) {
     const themePath = path.join(this._themesDir, themeName);
     if (!fs.existsSync(themePath)) {
       throw new UserError(
         `Theme "${themeName}" not found within the "${this._themesDir}" folder`);
     }
+    const updateBranch = branch ? branch : 'master';
```
cea2aj, line 101: You can shorten this line to this: `const updateBranch = branch || 'master';`

src/commands/upgrade/themeupgrader.js
oshi97: ok sgtm!

src/commands/upgrade/themeupgrader.js
```diff
@@ -58,13 +64,26 @@ class ThemeUpgrader {
...
+          displayName: 'Branch of theme to upgrade to',
+          type: 'singleoption',
+          options: branches
         }
       }
     }
   }
 
+  async _getThemeBranches() {
+    const branchesGit = simpleGit(
```
nanhu95, line 78: Creating a new instance of `simpleGit` here to be able to run `git branch --remote` with the appropriate repo.

src/commands/upgrade/themeupgrader.js
```diff
@@ -58,13 +64,26 @@ class ThemeUpgrader {
...
+        branch: {
+          displayName: 'Branch of theme to upgrade to',
+          type: 'singleoption',
+          options: branches
         }
       }
     }
   }
 
+  async _getThemeBranches() {
```
nanhu95, line 77: Due to `branch` returning a promise, this means upgrade's `describe` needs to be async, and thus the describe command has to be able to handle promises. This results in a chain of functions in the call stack getting turned into async functions and needing to handle this change. Open to suggestions on how best to handle this. Some brief thoughts: one way to mitigate this is to not list the branches in describe, or to use a git library that doesn't wrap everything as promises. cc @oshi97 @tmeyer2115

src/commands/upgrade/themeupgrader.js
oshi97: do we need to manually set the binary?

src/commands/upgrade/themeupgrader.js
```diff
@@ -75,24 +94,22 @@ class ThemeUpgrader {
 
   /**
    * Upgrades the given theme to the latest version.
-   * @param {string} themeName
-   * @param {boolean} disableScript
-   * @param {booolean} isLegacy
+   * @param {any} args The arguments provided to the upgrade command.
    */
-  async _upgrade(themeName, disableScript, isLegacy) {
+  async _upgrade(themeName, args) {
```
oshi97, line 99: would having `{ themeName, disableScript, isLegacy, branch }` instead of args be a little cleaner? Or maybe doing a `const { themeName, disableScript, isLegacy, branch } = args` on the first line? Could we also have jsdoc for the options inside args?

src/commands/upgrade/themeupgrader.js
oshi97: do we still need to say (defaults to master) here? we're not providing any defaulting behavior anymore

src/commands/upgrade/themeupgrader.js
```diff
@@ -128,21 +145,28 @@ class ThemeUpgrader {
...
    * @param {string} themeName
    * @param {string} themePath
+   * @param {string} branch
    */
-  async _recloneTheme(themeName, themePath) {
+  async _recloneTheme(themeName, themePath, branch) {
     await fs.remove(themePath);
     const themeRepoURL = getRepoForTheme(themeName);
-    await git.clone(themeRepoURL, themePath);
+    const updateBranch = branch || 'master';
```
oshi97, line 168: @tmeyer2115 I thought it would make sense that if somebody did a `jambo upgrade` without specifying --branch, that if they were already on a specific branch of the theme (lets say develop) it would upgrade to the newest version of develop instead of master. Realizing now that this is fine if the theme is a submodule, since the submodule keep track of which branch they're on, but not if it was imported as regular files. Would it make sense to keep track of the theme branch somewhere in the jambo.json? Alternatively we can just always upgrade to master if you don't specify a --branch

src/commands/upgrade/themeupgrader.js
tmeyer2115: Is it an issue for `describe` to have an `async execute()` method? There's at least one other command (`import`) whose `execute` is `async` (although it's not explicitly marked as such). If we just have to update some signatures to include `async`, I don't think that's a big deal. Are there other issues besides updating a number of signatures?

src/commands/upgrade/themeupgrader.js
```diff
@@ -58,36 +64,53 @@ class ThemeUpgrader {
...
 
   /**
    * Upgrades the given theme to the latest version.
-   * @param {string} themeName
-   * @param {boolean} disableScript
-   * @param {booolean} isLegacy
+   * @param {string} themeName The name of the theme
+   * @param {boolean} disableScript Whether to run the upgrade script
+   * @param {boolean} isLegacy Whether to use the isLegacy flag in the upgrade script
+   * @param {string} branch The nameo of the branch to upgrade to
```
tmeyer2115, line 103: "The name of the branch"

src/commands/upgrade/themeupgrader.js
tmeyer2115: I think, eventually, we may want the `import` command to take an optional branch as well. Without that, I don't think being on a non-master branch pre-`upgrade` is a common use case. So, I'd say we don't worry about it too much for now. If no `branch` is provided, and you didn't use submodules, you get `master` in the `upgrade`. When we update the `import` command, we can make `upgrade` more intelligent in figuring out the default `branch` for non-submodules.

nanhu95: I've created an [item](https://yexttest.atlassian.net/browse/SLAP-815) to track the work for `import` taking in a branch name as well. If we're okay with how the current upgrade logic works, then I'll keep these changes.

src/commands/upgrade/themeupgrader.js
nanhu95: Drawing on my knowledge, I don't think there's really any issue beyong having to update method signatures and adding in handling for both promises and normal values when getting command descriptions in `describe`. I was more curious if there could be a better way for handling the needed changes. If the currently implementation is okay, then I'm good with keeping it as is.

src/commands/upgrade/themeupgrader.js
nanhu95: Create this [item](https://yexttest.atlassian.net/browse/SLAP-816) to track the work!

src/commands/upgrade/themeupgrader.js
tmeyer2115: Okay! If there's a library that gets this in a synchronous way, we could certainly add that. I'm not sure such a library exists since getting branches from remote would require a network call. There's already a command that's `async`, I'm not opposed to having `describe` by `async` as well. Even if we find a way to do this synchronously, I bet in the future we're gonna have some other command that needs network calls for its `describe`.

src/commands/upgrade/themeupgrader.js
tmeyer2115: Seems fine to me! I think we can update later as part of the `import` work. Seems like the submodule case already has this behavior, so we'd just need to figure out something for the regular files case.

src/commands/upgrade/themeupgrader.js
```diff
@@ -58,36 +64,53 @@ class ThemeUpgrader {
...
         }
       }
     }
   }
 
+  async _getThemeBranches() {
+    const branchesGit = simpleGit(
+      path.join(this._themesDir, this.jamboConfig.defaultTheme));
+    const branches = await branchesGit.branch(['--remote']);
+    return branches.all.map(branch => branch.replace('origin/', ''))
```
oshi97, line 81: do we need to do a startsWith check here or is a name like origin/origin or feature/origin illegal

src/commands/describe/describecommand.js
```diff
@@ -41,11 +41,22 @@ module.exports = class DescribeCommand {
...
+          if (command.getAlias() !== 'describe') {
+            descriptions[command.getAlias()] = describeValue;
+          }
+        }
       }
-    }
-    return descriptions;
+    );
+    return Promise.all(describePromises).then(
+      () => descriptions
```
oshi97, line 59: super nit but can we have if formatted like
```js
blah.then(() => descriptions);
```

src/commands/upgrade/themeupgrader.js
nanhu95: Thanks for pointing this out! Based on git branch naming rules, it doesn't look like having a branch with `origin` in the name is illegal. I'll update the implementation to only strip out `origin/` at the beginning of the string.

oshi97 (approved):
lgtm!

PR #154, opened 2020-10-29
page cmd: add locales argument

nanhu95:
Adds a locales argument that can be passed into the page command.
The locales argument takes an array of locales, e.g. es, fr, it.

The page command's describe method returns an array of locales
found in config/locale_config.json.

J=SLAP-793
TEST=manual
When I run `jambo page` with the locales arg, such as
`jambo page --name test-locales --template vertical-standard --locales es fr`,
I see that es and fr configs are created for the test-locales in page in
addition to test-locales.json.
When I run `jambo describe` in a repo that has a locale_config.json
file with es and fr locales, I see that the page command returns
es and fr as options for the locales arg.

### DIFF
--- a/src/commands/page/add/pagecommand.js
+++ b/src/commands/page/add/pagecommand.js
@@ -34,12 +34,19 @@ class PageCommand {
         type: ArgumentType.STRING,
         description: 'template to use within theme',
         isRequired: false
+      }),
+      locales: new ArgumentMetadata({
+        type: ArgumentType.ARRAY,
+        itemType: ArgumentType.STRING,
+        description: 'locales the page should support',
+        isRequired: false
       })
     }
   }
 
   describe() {
     const pageTemplates = this._getPageTemplates();
+    const pageLocales = this._getPageLocales();
     return {
       displayName: 'Add Page',
       params: {
@@ -52,13 +59,18 @@ class PageCommand {
           displayName: 'Page Template',
           type: 'singleoption',
           options: pageTemplates
+        },
+        locales: {
+          displayName: 'Page Locales',
+          type: 'multioption',
+          options: pageLocales
         }
       }
     }
   }
 
   /**
-   * @returns {Array<string>}
+   * @returns {Array<string>} The page templates available in the theme
    */
   _getPageTemplates() {
     if (!this.defaultTheme || !this.themesDir) {
@@ -66,6 +78,28 @@ class PageCommand {
     }
     const pageTemplatesDir = path.resolve(this.themesDir, this.defaultTheme, 'templates');
     return fs.readdirSync(pageTemplatesDir);
+  }
+  
+  /**
+   * @returns {Array<string>} The locales that are configured in locale_configs.json
+   */
+  _getPageLocales() {
+    const configDir = this.jamboConfig.dirs.config;
+    const localeConfig = path.resolve(configDir, 'locale_config.json');
+    if (!configDir || ! localeConfig) {
+      return [];
+    }
+
+    const pageLocales = [];
+    const localeContentsRaw = fs.readFileSync(localeConfig);
+    const localeContentsJson = JSON.parse(localeContentsRaw);
+    for (const locale in localeContentsJson.localeConfig) {
+      // a page has 'en' locale by default, so it won't be listed as an option
+      if (locale !== 'en') {
+        pageLocales.push(locale);
+      }
+    }
+    return pageLocales;
   }
 
   execute(args) {
--- a/src/commands/page/add/pagecommand.js
+++ b/src/commands/page/add/pagecommand.js
@@ -46,7 +46,7 @@ class PageCommand {
 
   describe() {
     const pageTemplates = this._getPageTemplates();
-    const pageLocales = this._getPageLocales();
+    const pageLocales = this._getAdditionalPageLocales();
     return {
       displayName: 'Add Page',
       params: {
@@ -81,29 +81,32 @@ class PageCommand {
   }
   
   /**
-   * @returns {Array<string>} The locales that are configured in locale_config.json
+   * @returns {Array<string>} The additional locales that are configured in 
+   *                          locale_config.json
    */
-  _getPageLocales() {
+  _getAdditionalPageLocales() {
     const configDir = this.jamboConfig.dirs.config;
     if (!configDir) {
       return [];
     }
     const localeConfig = path.resolve(configDir, 'locale_config.json');
-    if (!localeConfig) {
+    try {
+      if (fs.existsSync(localeConfig)) {
+        const pageLocales = [];
+        const localeContentsRaw = fs.readFileSync(localeConfig);
+        const localeContentsJson = JSON.parse(localeContentsRaw);
+        const defaultLocale = localeContentsJson.default;
+        for (const locale in localeContentsJson.localeConfig) {
+          // don't list the default locale as an option
+          if (locale !== defaultLocale) {
+            pageLocales.push(locale);
+          }
+        }
+        return pageLocales;
+      }
+    } catch(err) {
       return [];
     }
-
-    const pageLocales = [];
-    const localeContentsRaw = fs.readFileSync(localeConfig);
-    const localeContentsJson = JSON.parse(localeContentsRaw);
-    const defaultLocale = localeContentsJson.default;
-    for (const locale in localeContentsJson.localeConfig) {
-      // don't list the default locale as an option
-      if (locale !== defaultLocale) {
-        pageLocales.push(locale);
-      }
-    }
-    return pageLocales;
   }
 
   execute(args) {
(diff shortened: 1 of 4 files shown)


src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,28 @@ class PageCommand {
     const pageTemplatesDir = path.resolve(this.themesDir, this.defaultTheme, 'templates');
     return fs.readdirSync(pageTemplatesDir);
   }
+  
+  /**
+   * @returns {Array<string>} The locales that are configured in locale_configs.json
+   */
```
cea2aj, line 85: nit: the file name is 'locale_config.json'

src/commands/page/add/pagecommand.js
```diff
@@ -34,12 +34,19 @@ class PageCommand {
         type: ArgumentType.STRING,
         description: 'template to use within theme',
         isRequired: false
+      }),
+      locales: new ArgumentMetadata({
+        type: ArgumentType.ARRAY,
+        itemType: ArgumentType.STRING,
+        description: 'locales the page should support',
```
cea2aj, line 41: What do you think of "locales that the page is generated for"?

cea2aj (approved):
Looks good!

src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,28 @@ class PageCommand {
     const pageTemplatesDir = path.resolve(this.themesDir, this.defaultTheme, 'templates');
     return fs.readdirSync(pageTemplatesDir);
   }
+  
+  /**
+   * @returns {Array<string>} The locales that are configured in locale_config.json
+   */
+  _getPageLocales() {
+    const configDir = this.jamboConfig.dirs.config;
+    const localeConfig = path.resolve(configDir, 'locale_config.json');
```
tmeyer2115, line 88: I think `path.resolve` will error if `configDir` is undefined here.

src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,28 @@ class PageCommand {
...
+    if (!configDir || ! localeConfig) {
+      return [];
```
tmeyer2115, line 90: This is a bit strange, but Jambo can support sites that don't have a `locale_config`, only a `global_config`. Prior to the internationalization work, sites only had a `global_config` (which listed a single locale). We want Jambo to still be compatible with these kinds of repos. Because of this, if no `locale_config` exists, you may need to check the `global_config.locale`.

src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,28 @@ class PageCommand {
...
+    }
+
+    const pageLocales = [];
+    const localeContentsRaw = fs.readFileSync(localeConfig);
+    const localeContentsJson = JSON.parse(localeContentsRaw);
+    for (const locale in localeContentsJson.localeConfig) {
+      // a page has 'en' locale by default, so it won't be listed as an option
```
tmeyer2115, line 97: Do we not include the `defaultLocale` among the list of options? I think we should, it may confuse people not to see it listed.

tmeyer2115: We might need to discuss how or if to display `defaultLocale` as an option with Sumo. cc: @mancusi

nanhu95: I was thinking that if we listed the `defaultLocale` in the options, a user may try to do something like `jambo page --locales en` when `en` is the default, and we would end up with a `[page].en.json` config. That being said, we could change the implementation to check against the `defaultLocale` being provided as an arg and ignore it if present.

src/commands/page/add/pagecommand.js
nanhu95: Is there a good example of what a `globalconfig.json` with a locale specified in it looks like? Also, related to the comment below about including the default locale in the list of options, if we end up deciding to not include the default, should we include `global_config.locale` if there's only one option there?

src/commands/page/add/pagecommand.js
mancusi: The only things we display to the user are what are in the options array so it'd need to be included to show in the UI.

src/commands/page/add/pagecommand.js
tmeyer2115: For such a site, the `global_config.locale` would include a singe, default locale. It sounds like we're not displaying the default locale in the UI, so we don't need to worry about this!

src/commands/page/add/pagecommand.js
```diff
@@ -34,12 +34,19 @@ class PageCommand {
         type: ArgumentType.STRING,
         description: 'template to use within theme',
         isRequired: false
+      }),
+      locales: new ArgumentMetadata({
+        type: ArgumentType.ARRAY,
+        itemType: ArgumentType.STRING,
+        description: 'locales to generate the page for',
```
tmeyer2115, line 41: Maybe something like 'additional locales to generate the page for'? Here and in the `describe` command, we want to indicate that we always generate the page for the `defaultLocale`, this argument allows us to specify additional locales also.

nanhu95: Sounds good to me, and thanks for catching it!

src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,32 @@ class PageCommand {
     const pageTemplatesDir = path.resolve(this.themesDir, this.defaultTheme, 'templates');
     return fs.readdirSync(pageTemplatesDir);
   }
+  
+  /**
+   * @returns {Array<string>} The locales that are configured in locale_config.json
```
tmeyer2115, line 84: We may want to update this description (and potentially the method name) to indicate that the default is not included in the array.

src/commands/page/add/pagecommand.js
```diff
@@ -81,29 +81,32 @@ class PageCommand {
   }
   
   /**
-   * @returns {Array<string>} The locales that are configured in locale_config.json
+   * @returns {Array<string>} The additional locales that are configured in 
+   *                          locale_config.json
    */
-  _getPageLocales() {
+  _getAdditionalPageLocales() {
     const configDir = this.jamboConfig.dirs.config;
```
tmeyer2115, line 88: Will this be an issue if someone does a `describe` before an `init`?

src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,35 @@ class PageCommand {
...
+        const localeContentsJson = JSON.parse(localeContentsRaw);
+        const defaultLocale = localeContentsJson.default;
+        for (const locale in localeContentsJson.localeConfig) {
+          // don't list the default locale as an option
+          if (locale !== defaultLocale) {
+            pageLocales.push(locale);
+          }
+        }
+        return pageLocales;
+      }
```
cea2aj, line 106: If `localeConfig` doesn't exist and it doesn't throw an error, I don't think we return an empty array here (but I think we probably should).

src/commands/page/add/pagecommand.js
```diff
@@ -81,29 +81,32 @@ class PageCommand {
...
-
-    const pageLocales = [];
-    const localeContentsRaw = fs.readFileSync(localeConfig);
-    const localeContentsJson = JSON.parse(localeContentsRaw);
-    const defaultLocale = localeContentsJson.default;
-    for (const locale in localeContentsJson.localeConfig) {
-      // don't list the default locale as an option
-      if (locale !== defaultLocale) {
-        pageLocales.push(locale);
+    try {
```
tmeyer2115, line 93: I'm not sure we want to wrap this whole thing in a `try ... catch`. If `localeConfig` doesn't exist, we should definitely return an empty array. If it does exist though, but the `JSON.parse` failed, that would be an indication that their `locale_config.json` was not properly formatted. In that case, I think we'd want to throw a `UserError`. Is that the correct error type @cea2aj ?

cea2aj: Yep, it would be a UserError if the locale_config can't be parsed

src/commands/page/add/pagecommand.js
nanhu95: Yes. I'll add a check for `this.jamboConfig` before attempting to resolve `configDir`

src/commands/page/add/pagecommand.js
nanhu95: Will break up the `try/catch` and follow the flow proposed above.

src/commands/page/add/pagecommand.js
```diff
@@ -67,6 +79,43 @@ class PageCommand {
...
+      return [];
+    }
+
+    const pageLocales = [];
+    const localeContentsRaw = fs.readFileSync(localeConfig);
+    let localeContentsJson;
+    try {
+      localeContentsJson = JSON.parse(localeContentsRaw);
+    } catch(err) {
+      throw new UserError('locale_config.json is not formatted properly ', err.stack);
```
cea2aj, line 108: nit: If we want this wording to be a little more consistent with the rest of the repo, we could word it like: "Could not parse locale_config.json"

cea2aj (approved):

PR #157, opened 2020-11-04
page command: use comment-json to parse locale_config.json

nanhu95:
Makes a change to use the comment-json package's parse function
when parsing locales from locale_config.json.

We need to use the comment-json package because the default
imported locale_config.json contains comments, which the
built-in JSON.parse thinks is invalid.

TEST=manual
Ran `jambo describe` in a repo that has a locale_config.json with
comments and saw that no errors occurred and that the locales
were returned correctly.


alexisgrow (approved):
LGTM!

PR #158, opened 2020-11-04
commands: all methods except execute are static

nanhu95:
Changes all Jambo commands so that every method except `execute`
is static.

This change is so that the cli will not need to instantiate a
command unless it actually needs to execute it.

Also updates the command interface so that all methods except
execute are static.

J=SLAP-817
TEST=manual
Ran all jambo commands and saw that there were no changes in behavior.

### DIFF
--- a/src/commands/build/buildcommand.js
+++ b/src/commands/build/buildcommand.js
@@ -11,15 +11,15 @@ class BuildCommand {
     this.sitesGenerator = sitesGenerator;
   }
 
-  getAlias() {
+  static getAlias() {
     return 'build';
   }
 
-  getShortDescription() {
+  static getShortDescription() {
     return 'build the static pages for the site';
   }
 
-  args() {
+  static args() {
     return {
       jsonEnvVars: new ArgumentMetadata({
         type: ArgumentType.ARRAY,
@@ -30,7 +30,7 @@ class BuildCommand {
     }
   }
 
-  describe() {
+  static describe() {
     return {
       displayName: 'Build Pages'
     }
--- a/src/commands/build/buildcommand.js
+++ b/src/commands/build/buildcommand.js
@@ -1,6 +1,6 @@
-const { SitesGenerator } = require('./sitesgenerator');
 const { ArgumentMetadata, ArgumentType } = require('../../models/commands/argumentmetadata');
 const UserError = require('../../errors/usererror');
+const { isCustomError } = require('../../utils/errorutils');
 
 /**
  * BuildCommand builds all pages in the Jambo repo and places them in the
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -31,7 +31,6 @@ class CommandImporter {
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
         .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
         .map(require)
-        .map(commandFn => commandFn(this._jamboConfig));
 
       // Remove the merged commands directory from 'public' as it is no longer needed.
       fs.removeSync(mergedDirectory);
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -5,10 +5,9 @@ const fs = require('fs-extra');
  * Imports all custom {@link Command}s within a Jambo repository.
  */
 class CommandImporter {
-  constructor(outputDir, jamboConfig, themeDir) {
+  constructor(outputDir, themeDir) {
     this._outputDir = outputDir;
     this._themeDir = themeDir;
-    this._jamboConfig = jamboConfig;
   }
 
   /**
--- a/src/commands/commandregistry.js
+++ b/src/commands/commandregistry.js
@@ -19,10 +19,10 @@ class CommandRegistry {
   /**
    * Registers a new {@link Command} class with the CLI.
    *
-   * @param {Class} commandClazz 
+   * @param {Class} commandClass 
    */
-  addCommand(commandClazz) {
-    this._commandsByName[commandClazz.getAlias()] = commandClazz;
+  addCommand(commandClass) {
+    this._commandsByName[commandClass.getAlias()] = commandClass;
   }
 
   /**
--- a/src/commands/commandregistry.js
+++ b/src/commands/commandregistry.js
@@ -1,8 +1,6 @@
 const InitCommand = require('../commands/init/initcommand');
-const PageScaffolder = require('./page/add/pagescaffolder');
 const PageCommand = require('./page/add/pagecommand');
 const OverrideCommand = require('./override/overridecommand');
-const SitesGenerator = require('./build/sitesgenerator');
 const BuildCommand = require('./build/buildcommand');
 const DescribeCommand = require('../commands/describe/describecommand');
 const JamboTranslationExtractor = require('./extract-translations/jambotranslationextractor');
@@ -21,11 +19,10 @@ class CommandRegistry {
   /**
    * Registers a new {@link Command} with the CLI.
    *
-   * @param {Object} command an object containing a command's class and a way
-   *                         to instantiate the class
+   * @param {Command} command 
    */
   addCommand(command) {
-    this._commandsByName[command.clazz.getAlias()] = command;
+    this._commandsByName[command.getAlias()] = command;
   }
 
   /**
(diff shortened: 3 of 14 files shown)


src/commands/commandregistry.js
```diff
@@ -50,27 +50,43 @@ class CommandRegistry {
...
-      [ initCommand.getAlias() ]: initCommand,
-      [ importCommand.getAlias() ]: importCommand,
-      [ pageCommand.getAlias() ]: pageCommand,
-      [ overrideCommand.getAlias() ]: overrideCommand,
-      [ buildCommand.getAlias() ]: buildCommand,
-      [ upgradeCommand.getAlias() ]: upgradeCommand,
-      [ describeCommand.getAlias() ]: describeCommand,
-      [ extractTranslationsCommand.getAlias() ]: extractTranslationsCommand
+      [ InitCommand.getAlias() ]: {
+        obj: InitCommand,
```
nanhu95, line 57: `class` is a reserved keyword, so using `obj` here. I don't really like the naming, so open to suggestions!

src/commands/commandimporter.js
```diff
@@ -31,7 +31,6 @@ class CommandImporter {
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
         .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
         .map(require)
-        .map(commandFn => commandFn(this._jamboConfig));
```
tmeyer2115, line 34: Without this, do we need `this._jamboConfig` in the class any longer?

src/commands/commandregistry.js
```diff
@@ -25,7 +25,7 @@ class CommandRegistry {
    * @param {Command} command
```
tmeyer2115, line 25: Update JSDoc here

src/commands/commandregistry.js
```diff
@@ -25,7 +25,7 @@ class CommandRegistry {
...
    */
   addCommand(command) {
-    this._commandsByName[command.getAlias()] = command;
+    this._commandsByName[command.obj.getAlias()] = command;
```
tmeyer2115, line 28: We might want to change the name of `_commandsByName` as the map no longer contains a command. In Java, the word `class` is also reserved. In situations like these, Java programmers use the word `clazz`. Maybe we can adopt that convention here as well. It's also possible JS may have it's own convention instead of `clazz`.

src/commands/commandregistry.js
```diff
@@ -50,27 +50,43 @@ class CommandRegistry {
...
-      [ importCommand.getAlias() ]: importCommand,
-      [ pageCommand.getAlias() ]: pageCommand,
-      [ overrideCommand.getAlias() ]: overrideCommand,
-      [ buildCommand.getAlias() ]: buildCommand,
-      [ upgradeCommand.getAlias() ]: upgradeCommand,
-      [ describeCommand.getAlias() ]: describeCommand,
-      [ extractTranslationsCommand.getAlias() ]: extractTranslationsCommand
+      [ InitCommand.getAlias() ]: {
+        obj: InitCommand, 
+        constructor: () => new InitCommand()
```
tmeyer2115, line 58: Minor nit, but maybe we call this `factory`? This isn't the same thing as the classes constructor, so it may be confusing to name it that way.

src/commands/commandregistry.js
```diff
@@ -25,7 +25,7 @@ class CommandRegistry {
    * @param {Command} command
    */
   addCommand(command) {
```
tmeyer2115, line 27: Should we call the methods of this class `addCommand` and `getCommands`? The registry no longer deals in actual `Command`s.

src/commands/import/themeimporter.js
```diff
@@ -45,7 +45,7 @@ class ThemeImporter{
     }
   }
 
-  describe() {
+  static describe() {
```
tmeyer2115, line 48: This is a bit outside the scope of this item, but do we have unit tests for the describe command? If so, can we add a unit test simulating the describe command being run in an empty Jambo repo?

src/yargsfactory.js
```diff
@@ -28,11 +29,13 @@ class YargsFactory {
    * @returns {Object<string, ?>} The {@link yargs} CommandModule for the {@link Command}.
    */
   _createCommandModule(command) {
```
tmeyer2115, line 31: Can we update the param name? It's not really taking in a `Command` any more

src/yargsfactory.js
```diff
@@ -28,11 +29,13 @@ class YargsFactory {
...
+    const commandClass = command.obj;
+    const commandInstance = command.constructor(this._jamboConfig);
```
tmeyer2115, line 33: By having the `commandInstance` constructed here, we're not waiting until the command is actually invoked. I think this would put us in the same situation we're in currently. Regardless of which command you run, we instantiate all of them.

src/commands/commandimporter.js
nanhu95: Good catch, will remove!

src/commands/commandregistry.js
nanhu95: I think we should have come up with a term to describe the object containing a command's class and the way to instantiate it and use that terminology in naming. Maybe something like `commandCapsule`? Open to naming suggestions.

Based on some searching it doesn't seem like JS has a preferred covention for substituting `class`. Will switch to using `clazz`.

src/commands/import/themeimporter.js
nanhu95: There are unit tests for the command. Currently the test is run without an actual jambo config (see changes [here](https://github.com/yext/jambo/pull/158/files#diff-a728e94d3f64baa7c359b34c25ef633add5b64589f89c7268fe71207d550739cL4) made in this PR for the test), is this what you are thinking of?

tmeyer2115: I want to make sure that every built-in command's `describe` can gracefully handle an empty repo. I guess that would require adding a test to each command's suite. I can make an item for that if those tests don't exist!

nanhu95: Gotcha. We don't currently have that set up. I can make the item right now!

src/commands/commandregistry.js
nanhu95: Had a discussion with Tom wrt implementation approach and we will now do the following: custom commands only need to export the class (no need to export a function that instantiates the class anymore), and in commandregistry.js, we will just return a dictionary of command name to command class (uninstantiated); we will move the logic of instantiating the class into yargsfactory.js. This should simplify our contract for custom commands (the only dependency we can provide them is the jambo config) and also remove the complexities we're introducing into commandregistry.js with the need to return two things instead of one. This also eliminates the need to change around naming/terminology in variable and method names.

src/commands/build/buildcommand.js
```diff
@@ -1,6 +1,6 @@
-const { SitesGenerator } = require('./sitesgenerator');
 const { ArgumentMetadata, ArgumentType } = require('../../models/commands/argumentmetadata');
 const UserError = require('../../errors/usererror');
+const { isCustomError } = require('../../utils/errorutils');
```
tmeyer2115, line 3: Do we use `isCustomError` in this class?

src/commands/commandregistry.js
```diff
@@ -21,8 +19,7 @@ class CommandRegistry {
   /**
    * Registers a new {@link Command} with the CLI.
    *
-   * @param {string} name
-   * @param {Command} command
+   * @param {Command} command
```
tmeyer2115, line 22: We need to make sure that the JSDoc and param names are correct and informative. We're now passing in the class for a `Command`, instead of a `Command` instance. Potentially, we could rename `command` to `commandClazz`. 

There may be other methods in this class that need param names and JSDoc updated as well.

src/yargsfactory.js
```diff
@@ -43,7 +47,29 @@ class YargsFactory {
             });
         });
       },
-      handler: argv => command.execute(argv)
+      handler: argv => {
+        const commandName = command.name;
+        let commandInstance;
+        switch (commandName) {
```
tmeyer2115, line 53: Let's factor this `switch` statement out to it's own method.

src/yargsfactory.js
```diff
@@ -43,7 +47,29 @@ class YargsFactory {
...
+          case 'DescribeCommand':
+            commandInstance = new command(
```
tmeyer2115, line 55: Is `command` an object? Above, it looks like we're accessing `command.name`. If it's an object, I'm not sure that `new command` will work.

src/commands/build/buildcommand.js
nanhu95: Yes we use it in `execute`.

tmeyer2115: Gotcha! Guess we missed adding it earlier. Was that causing things to break?

src/yargsfactory.js
nanhu95: `command` is recognized as a class and based on my testing, it looks like `new command` is working. The describe, page, and build commands all work as expected when run.

src/commands/build/buildcommand.js
nanhu95: It would only have caused things to break only if `jambo build` couldn't execute successfully. For context, I noticed this in the process of testing when I ran `jambo build` without passing in a SitesGenerator properly and caused it to fail.

src/yargsfactory.js
tmeyer2115: Ah, it looks like ES6 classes may have a special `name` attribute. Can we call it `commandClazz` or something like that to differentiate between it and an actual `Command` instance?

src/yargsfactory.js
```diff
@@ -14,25 +18,25 @@ class YargsFactory {
...
+      cli.command(this._createCommandModule(commandClazz));
     });
     cli.strict()
 
     return cli;
   }
 
   /**
-   * @param {Command} command A Jambo {@link Command}.
+   * @param {Class} commandClazz A Jambo {@link Command}'s class'.
```
tmeyer2115, line 31: "{@link Command}'s class."

src/yargsfactory.js
nanhu95: Yep if you look at the latest commit, I've changed the naming from `command` -> `commandClazz`

src/yargsfactory.js
```diff
@@ -43,8 +47,40 @@ class YargsFactory {
             });
         });
       },
-      handler: argv => command.execute(argv)
+      handler: argv => {
+        const commandName = commandClazz.name;
+        const commandInstance = this._getCommandInstance(commandName, commandClazz);
```
tmeyer2115, line 52: Do we need to pass in `commandName` here? It's already encapsulated in the `commandClazz`.

src/yargsfactory.js
```diff
@@ -43,8 +47,40 @@ class YargsFactory {
...
+  }
+
+  /**
+   * @param {String} commandName the name of the Jambo command's class
+   * @param {Class} commandCLazz the class of the Jambo command
+   * @returns {Command} the instantiated Jambo command
+   */
+  _getCommandInstance(commandName, commandClazz) {
+    let commandInstance;
+    switch (commandName) {
```
tmeyer2115, line 65: Minor nit: Can we call this `className` or something like that? The term `commandName` sounds a good bit like the alias. I totally didn't realize this before, but we could potentially call the param `commandClass`, which isn't reserved. Take a look to see if JS/Java convention dictates we call it `commandClazz` or `commandClass`.

src/yargsfactory.js
nanhu95: Good catch, we don't need to! Will make the change

src/yargsfactory.js
nanhu95: It doesn't look like there's much convention for when "class" is part of a variable name (as opposed to when it's standalone). I think `commandClass` is easier to read, will make the change.

src/yargsfactory.js
```diff
@@ -48,37 +48,36 @@ class YargsFactory {
...
+        const commandInstance = this._createCommandInstance(commandClass);
         commandInstance.execute(argv);
       }
     }
   }
 
   /**
-   * @param {String} commandName the name of the Jambo command's class
-   * @param {Class} commandCLazz the class of the Jambo command
+   * @param {Class} commandCLass the class of the Jambo command
```
tmeyer2115, line 58: `commandClass`

tmeyer2115 (commented):
One small thing, then lgtm!

src/yargsfactory.js
nanhu95: Thanks for the catch!

tmeyer2115 (approved):

PR #159, opened 2020-11-10
Ensure that the 'describe' command outputs JSON

tmeyer2115:
The `describe` command was logging a JS object to the console. This format was not parsable by SGS. This PR updates `describe` so that JSON is logged instead. 

J=SLAP-854
TEST=auto, manual

Updated the `describe` unit tests. Also ran the command on a test repo and saw the expected output.

### DIFF
--- a/src/commands/describe/describecommand.js
+++ b/src/commands/describe/describecommand.js
@@ -31,10 +31,8 @@ class DescribeCommand {
   }
 
   async execute() {
-    console.dir(await this._getCommandDescriptions(), {
-      depth: null,
-      maxArrayLength: null
-    });
+    const descriptions = await this._getCommandDescriptions();
+    console.log(JSON.stringify(descriptions));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/commands/describe/describecommand.js
```diff
@@ -31,10 +31,8 @@ class DescribeCommand {
   }
 
   async execute() {
-    console.dir(await this._getCommandDescriptions(), {
-      depth: null,
-      maxArrayLength: null
-    });
+    const descriptions = await this._getCommandDescriptions();
+    console.log(JSON.stringify(descriptions));
```
oshi97, line 35: do we need to do something like JSON.stringify(descriptions, null, 2) for it to be human readable, or do we only care about it being parsed out of the console?

nanhu95:
It looks like the `describecommand` test is failing.

src/commands/describe/describecommand.js
tmeyer2115: Ideally, it's human readable too! Updated.

nanhu95 (approved):
LGTM!

mancusi (approved):
lgtm!

oshi97 (approved):

alexisgrow (approved):
LG!

PR #163, opened 2020-11-16
add isAbsoluteUrl helper

oshi97:
This is needed in the theme. When custom hbs helpers are implemented,
this can be cut over.

J=SLAP-861
TEST=manual

test that urls like //yext.com will still be recognized as absolute
test using a data:image/gif;base64, url for favicon

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -310,6 +310,15 @@ class SitesGenerator {
       return str && str.match(regex);
     });
 
+    /**
+     * Determine whether a URL is absolute or not.
+     * Common examples: "mailto:slapshot@gmail.com", "//yext.com", "https://yext.com"
+     */
+    hbs.registerHelper('isAbsoluteUrl', function(str) {
+      const absoluteURLRegex = /(\/\/|^[a-zA-Z]+:)/;
+      return str && str.match(absoluteURLRegex);
+    });
+
     hbs.registerHelper('all', function(...args) {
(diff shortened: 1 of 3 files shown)


src/commands/build/sitesgenerator.js
```diff
@@ -310,6 +310,15 @@ class SitesGenerator {
       return str && str.match(regex);
     });
 
+    /**
+     * Determine whether a URL is absolute or not.
+     * Common examples: "mailto:slapshot@gmail.com", "//yext.com", "https://yext.com"
+     */
+    hbs.registerHelper('isAbsoluteUrl', function(str) {
+      const absoluteURLRegex = /(\/\/|^[a-zA-Z]+:)/;
```
tmeyer2115, line 318: Would this be more robust if it was the negation of `/^(\.\.)/`? Or is that too permissive?

tmeyer2115: Actually, it's probably better to be explicit about what an absolute URL should look like. Simply saying: "it's anything not of this form" feels like it could get us into trouble.

tmeyer2115, line 318: Playing around with my command line, it looks like `..//` has the same effect as `../`. So, it may be possible that someone uses `//` in a relative URL? The regex may need to be more explicit.

oshi97: ah it looks like ..// is also a relative url in the browser as well, will update

tmeyer2115 (approved):

PR #165, opened 2020-11-20
Recognize urls that start with only 1 slash as absolute

oshi97:
This commit updates the isAbsoluteUrl helper to recognize
urls like "/my-img.svg" as absolute urls. On a random website,
say yext.com, that url will point to yext.com/my-img.svg no matter
what page it is on.

Theme PR this is needed here
https://github.com/yext/answers-hitchhiker-theme/pull/482
will upgrade jambo ver once this is merged in

T=https://yextops.zendesk.com/agent/tickets/366842
J=SLAP-881
TEST=manual

test that the WHO site can specify an iconUrl that points to a static
asset

tested that a fresh site can also specify an iconUrl in a CTA
that points to a static asset, and also can directly put in
img tags with a static asset src

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "jambo",
-  "version": "1.9.0",
+  "version": "1.9.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


alexisgrow (approved):
this lgtm!

PR #167, opened 2020-11-23
Only add custom commands to the CLI instance when necessary.

tmeyer2115:
Previously, Jambo would always import custom commands, even if they weren't
necessary. Importing these commands and adding them to the CLI is only required
if `--help`, `describe`, or a command not part of the built-in set, is run.
This PR updates the `cli.js` file to only import and add custom commands to the
registry under these three circumstances.

I also took this oppurtunity to refactor the `cli.js` file. There are now
methods for each step of the CLI initialization process.

J=SLAP-888
TEST=manual

Tested these scenarios:

- Invoked Jambo without adding an actual command.
- Invoked `--help` and saw the expected output. This was done in an initialized
  and uninitialized repo.
- Invoked `describe` and saw the expected output. This was done in an
  initialized and uninitialized repo.
- Invoked a couple of built-in commands and verified they worked as expected.
- Invoked a custom command and verified it worked as expected.
- Used `--help` on both a built-in and custom command and verified that it
  worked as expected.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -17,10 +17,56 @@ process.on('uncaughtException', err => {
   exitWithError(err);
 });
 
-const jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
+buildJamboCLI();
 
-const commandRegistry = new CommandRegistry(jamboConfig);
-if (jamboConfig && jamboConfig.dirs && jamboConfig.dirs.output) {
+/**
+ * @returns {Object} A fully built Jambo CLI instance.
+ */
+function buildJamboCLI() {
+  const jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
+  const commandRegistry = new CommandRegistry();
+
+  if (process.argv.length < 3) {
+    console.error('You must provide Jambo with a command.');
+    return;
+  }
+
+  const shouldParseCustomCommands = 
+    shouldImportCustomCommands(process.argv[2], commandRegistry);
+  const canParseCustomCommands = 
+    jamboConfig && jamboConfig.dirs && jamboConfig.dirs.output;
+  
+  if (shouldParseCustomCommands && canParseCustomCommands) {
+    importCustomCommands(jamboConfig, commandRegistry);
+  }
+  
+  const yargsFactory = new YargsFactory(commandRegistry, jamboConfig);
+  const options = yargsFactory.createCLI();
+  return options.argv;
+}
+
+/**
+ * Determines if custom {@link Command}s should be imported and added to the CLI instance.
+ * 
+ * @param {string} invokedCommand The Jambo command that was invoked from the 
+ *                                command line.
+ * @param {CommandRegistry} commandRegistry The registry containing all built-in commands.
+ * @returns {boolean} If custom {@link Command}s need to be added to the CLI instance.    
+ */
+function shouldImportCustomCommands(invokedCommand, commandRegistry) {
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -17,10 +17,56 @@ process.on('uncaughtException', err => {
...
+function buildJamboCLI() {
+  const jamboConfig = fs.existsSync('jambo.json') && parseJamboConfig();
+  const commandRegistry = new CommandRegistry();
+
+  if (process.argv.length < 3) {
+    console.error('You must provide Jambo with a command.');
+    return;
+  }
+
+  const shouldParseCustomCommands =
```
nanhu95, line 34: (nit) could we have the variable name and the function named be aligned?

tmeyer2115: Ah, so in the Jambo project, a tab is only two spaces. The convention, when breaking on an equals sign, is to make a newline and indent one tab further than the previous line.

src/cli.js
```diff
@@ -17,10 +17,56 @@ process.on('uncaughtException', err => {
...
+
+  if (process.argv.length < 3) {
+    console.error('You must provide Jambo with a command.');
+    return;
+  }
+
+  const shouldParseCustomCommands = 
+    shouldImportCustomCommands(process.argv[2], commandRegistry);
+  const canParseCustomCommands = 
+    jamboConfig && jamboConfig.dirs && jamboConfig.dirs.output;
```
alexisgrow, line 37: Just for my own knowledge, why is the output directory required to parse custom commands?

src/cli.js
```diff
@@ -17,10 +17,56 @@ process.on('uncaughtException', err => {
...
+ * 
+ * @param {string} invokedCommand The Jambo command that was invoked from the 
+ *                                command line.
+ * @param {CommandRegistry} commandRegistry The registry containing all built-in commands.
+ * @returns {boolean} If custom {@link Command}s need to be added to the CLI instance.    
+ */
+function shouldImportCustomCommands(invokedCommand, commandRegistry) {
+  return invokedCommand === '--help' ||
+    invokedCommand === 'describe' ||
+    !commandRegistry.getAliases().includes(invokedCommand);
```
alexisgrow, line 59: Do we want to print a warning if the command alias is already in use?

alexisgrow (approved):
LGTM, two minor questions

src/cli.js
nanhu95: Gotcha! Sorry I originally meant could we have the name be the same for the variable and the function (so `shouldParseCustomCommands` for both, or `shouldImportCustomCommands` for both)

src/cli.js
tmeyer2115: This is a not so great piece of the design. It should be fixed sooner rather than later, but isn't as urgent as this issue. The output directory is needed because we merge the Theme's commands directory with the top-level commands directory. We do this merge operation in the output directory (just like static). This is how we let people override the Theme's custom commands.

src/cli.js
tmeyer2115: The way things work now, we never check to see if they've got an override of a built-in command, we immediately invoke the built-in one. If the command has the alias of a built-in one, we don't want to parse the custom files at all. As such, we won't do any collision detection. We decided on Friday this was the desired behavior. If they want to override a built-in command, they can make a custom command but just name it differently. When we write up docs for creating custom commands, we should make sure to make this behavior clear! I can make a note for that.

PR #168, opened 2020-11-23
Add specific logic to construct Array-type options in the CLI.

tmeyer2115:
Previously, every type of option was added to Yargs using `yargs.option`. But,
Yargs has special, different syntax for registering Array-type options. This
PR ensures we uses that syntax whenever `ArgumentMetadata.getType` is `ARRAY`.

TEST=manual

Tested the following scenarios:
- An array-type option is added to Yargs correctly.
- A string-type option is still added to Yargs correctly.
- When no explicit default is specified, an Array-type option defaults to `[]`.


nanhu95 (approved):
LGTM!

PR #169, opened 2020-11-23
Correct logic for determining if custom commands should be imported.

tmeyer2115:
The logic for determining if custom commands should be imported had a small
bug. When `jambo --version` was invoked, Jambo attempted to import custom
commands and register them with the CLI. This behavior was incorrect.

TEST=manual

Made sure that `--version` did not cause custom commands to be imported and
added to the registry.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "jambo",
-  "version": "1.9.1",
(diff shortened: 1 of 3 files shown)


nanhu95 (approved):
LGTM!

PR #171, opened 2020-12-01
rename hbs helper isAbsoluteUrl to isNonRelativeUrl

oshi97:
The helper also matches on root relative links,
which are technically not "absolute"


creotutar (approved):
👍 👍 👍 LGTM

PR #172, opened 2020-12-01
Revert import --addAsSubmodule param name change

oshi97:
It looks like this param was accidentally changed to
--addThemeAsSubmodule some time ago, this commit
reverts this change so that theme importer is consistent
with its describe output (and also the theme sync script).

TEST=manual

try importing a theme with --addAsSubmodule false, no error
try importing a theme with --addThemeAsSubmodule false, jambo
complains about unknown argument


tmeyer2115 (approved):

nanhu95 (approved):
LGTM!

PR #174, opened 2020-12-04
Add support for theme hbshelpers

oshi97:
This commit adds support for an hbshelpers
folder in the theme, which contains custom
hbs helpers used in the theme, with one
helper per file, with the helper name
being the name of the file.

The babel and isNonRelativeUrl helpers were
not removed from jambo for when CBD
npm upgrades a theme's jambo version
without upgrading the theme itself, which
would cause those helpers to be completing missing.
The theme can still override these helpers if changes
need to be made.

J-SLAP-894
TEST=manual, auto

test that, after removing the babel and isNonRelativeUrl
helpers from jambo, and adding them to the themes hbshelpers
folder, that I can still build the page and it works in ie11

test that jambo build will work even if the hbshelpers folder is missing
(but will complain about missing helpers if there are any 
undefined helpers)

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -17,6 +17,7 @@ const PartialsRegistry = require('../../models/partialsregistry');
 const HandlebarsPreprocessor = require('../../handlebars/handlebarspreprocessor');
 const { stripExtension, isValidFile, getPageName } = require('../../utils/fileutils');
 const registerHbsHelpers = require('../../handlebars/registerhbshelpers');
+const registerCustomHbsHelpers = require('../../handlebars/registercustomhbshelpers');
 const SystemError = require('../../errors/systemerror');
 const Translator = require('../../i18n/translator/translator');
 const UserError = require('../../errors/usererror');
@@ -130,12 +131,24 @@ class SitesGenerator {
     const translations = 
       await this._extractTranslations(locales, configRegistry.getLocalizationConfig());
 
-    // Register needed Handlebars helpers.
+    // Register built-in Jambo Handlebars helpers.
     console.log('Registering Jambo Handlebars helpers');
     try {
       registerHbsHelpers(hbs);
     } catch (err) {
       throw new SystemError('Failed to register jambo handlebars helpers', err.stack);
+    }
+
+    // Register the default theme's Handlebars helpers from the hbshelpers folder.
+    const customHbsHelpersDir =
+      path.resolve(config.dirs.themes, config.defaultTheme, 'hbshelpers')
+    if (fs.existsSync(customHbsHelpersDir)) {
+      try {
+        console.log('Registering custom Handlebars helpers from the default theme');
+        registerCustomHbsHelpers(hbs, customHbsHelpersDir);
+      } catch (err) {
+        throw new UserError('Failed to register custom handlebars helpers', err.stack);
+      }
     }
 
     const pageSets = GENERATED_DATA.getPageSets();
--- a/src/handlebars/registercustomhbshelpers.js
+++ b/src/handlebars/registercustomhbshelpers.js
@@ -1,0 +1,22 @@
+const fs = require('file-system');
+const UserError = require('../errors/usererror');
+const { stripExtension, isValidFile } = require('../utils/fileutils');
+
+/**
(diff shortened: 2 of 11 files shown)


src/commands/build/sitesgenerator.js
```diff
@@ -130,14 +131,26 @@ class SitesGenerator {
...
     } catch (err) {
       throw new SystemError('Failed to register jambo handlebars helpers', err.stack);
     }
 
+    // Register the default theme's Handlebars helpers from the hbshelpers folder.
+    const customHbsHelpersDir =
+      path.resolve(config.dirs.themes, config.defaultTheme, 'hbshelpers')
+    if (fs.existsSync(customHbsHelpersDir)) {
+      try {
+        console.log('Registering custom Handlebars helpers from the default theme');
```
creotutar, line 147: very small nit: do we wanna use `console.info` here?

oshi97: I didn't know about that one! we do console.log everywhere else in sitesgenerator, and apparently console.info is an alias for console.log, so probably not, but was there some reason you had in mind?

src/handlebars/registercustomhbshelpers.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('file-system');
+const UserError = require('../errors/usererror');
+const { stripExtension, isValidFile } = require('../utils/fileutils');
+
+/**
+ * Register's Jambo's built-in hbs helpers
```
tmeyer2115, line 6: Update JSDoc. Also, we should have a new line in between this and the `@params`.

src/handlebars/registercustomhbshelpers.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {Handlebars} hbs the handlebars instance
+ * @param {string} pathToCustomHelpers the path to the custom hbs helpers directory
+ */
+module.exports = function registerCustomHbsHelpers(hbs, pathToCustomHelpers) {
+  fs.recurseSync(pathToCustomHelpers, (path, relative, filename) => {
```
tmeyer2115, line 11: This is a bit different than the pattern for custom commands. With the custom command parsing, we look only for JS files in the top-level of the directory. We don't recurse down to any sub-folders. Our reasoning was that sub-folders could house some common functions, referenced by the commands. These functions wouldn't be commands themselves though. Would something like that make sense here for the helpers?

tests/fixtures/handlebars/customhelpers/isNonRelativeUrl.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Determine whether a URL is relative or not.
+ * Common examples: "mailto:slapshot@gmail.com", "//yext.com", "https://yext.com",
+ * "/my-img.svg"
+ */
+module.exports = function isNonRelativeUrl(str) {
```
tmeyer2115, line 6: I do think we should get the Babel helper out of Jambo at some point. I feel it's too much responsibility for Jambo and it seems a bit out of place. Theme's should specify how/if they want their code polyfilled. I recognize that this would cause issues in CBD though. Totally fine to leave it in Jambo for now, but I think we should come up with a plan to remove it long term.

src/commands/build/sitesgenerator.js
creotutar: ah just checked and node doesn't really make a difference in output for the two. in chrome it would display differently. i liked having `console.info` on logs that were meant to be there because people often use `console.log` to debug, would make it easier to cmd+f for unwanted logs

fine with it here as 1) not in browser, 2) used everywhere else

tests/fixtures/handlebars/customhelpers/isNonRelativeUrl.js
oshi97: Yeah definitely agree! For now, if the theme wants to make changes to the babel helper it can still be overridden so we shouldn't ever be stuck

src/handlebars/registercustomhbshelpers.js
oshi97: ah yeah that's reasonable! also good to be consistent

cea2aj:
Other than the outstanding comments, lgtm!

tmeyer2115 (approved):

PR #177, opened 2020-12-09
Add postimport hook, remove default postimport

oshi97:
This commit adds the postimport hook to jambo
import, and removes the default postimport
logic (which has been moved to the theme).
We can remove the logic in this specific case,
because anybody who calls jambo import will
be importing the newest version of the theme,
which will already have the postimport logic
moved into it.

This commit allows users to specify a postimport.js
(or any file extension, e.g. postimport.sh) script
to be run after an import.

This PR also updates the postupgrade script to have
the same capability (i.e. you can now specify an
upgrade.sh, instead of upgrade.js)

Finally it adds stdout, stderr, and error handling
to the commandexecutor, which might be renamed to
something like scriptexecutor in a separate PR,
now that we have custom jambo commands.

J=SLAP-833
TEST=manual

test that I can specify a postimport.sh or postimport.js
test that stdout, stderr, and errors are logged correctly
test that it works when import the theme as either a submodule
or checked in files
test that I can still upgrade the theme to a branch

### DIFF
--- a/src/commands/import/themeimporter.js
+++ b/src/commands/import/themeimporter.js
@@ -1,25 +1,24 @@
-const fs = require('file-system');
+const path = require('path');
 const simpleGit = require('simple-git/promise');
 const git = simpleGit();
-const {
-  stringify,
-  assign
-} = require('comment-json');
-const { ShadowConfiguration, ThemeShadower } = require('../override/themeshadower');
+const { ThemeShadower } = require('../override/themeshadower');
 const { getRepoForTheme } = require('../../utils/gitutils');
 const SystemError = require('../../errors/systemerror');
 const UserError = require('../../errors/usererror');
 const { isCustomError } = require('../../utils/errorutils');
-const { FileNames } = require('../../constants');
 const { ArgumentMetadata, ArgumentType } = require('../../models/commands/argumentmetadata');
+const { CustomCommand } = require('../../utils/customcommands/command');
+const { CustomCommandExecuter } = require('../../utils/customcommands/commandexecuter');
+const { searchDirectoryIgnoringExtensions } = require('../../utils/fileutils');
 
 /**
  * ThemeImporter imports a specified theme into the themes directory.
  */
-class ThemeImporter{
+class ThemeImporter {
   constructor(jamboConfig) {
     this.config = jamboConfig;
     this._themeShadower = new ThemeShadower(jamboConfig);
+    this._postImportHook = 'postimport';
   }
 
   static getAlias() {
@@ -38,7 +37,7 @@ class ThemeImporter{
         isRequired: true
       }),
       addAsSubmodule: new ArgumentMetadata({
-        type: ArgumentType.BOOLEAN, 
+        type: ArgumentType.BOOLEAN,
         description: 'import the theme as a submodule',
         defaultValue: true
       }),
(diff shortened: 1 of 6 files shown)


src/utils/fileutils.js
```diff
@@ -32,3 +35,28 @@ isValidFile = function(fileName) {
...
+ * of the first found.
+ * 
+ * @param {string} desiredFile
+ * @param {string} directoryPath
+ * @returns {string|undefined} the fileName, if it exists, otherwise undefined.
+ */
+searchDirectoryIgnoringExtensions = function(desiredFile, directoryPath) {
+  const dirEntries = fs.readdirSync(directoryPath);
+  for (const dirEntry of dirEntries) {
+    if (desiredFile !== stripExtension(dirEntry)) {
```
nanhu95, line 52: Thoughts on having the check be `if (desiredFile === stripExtension(dirEntry))` and just putting the logic in the `if`?

src/utils/fileutils.js
```diff
@@ -32,3 +35,28 @@ isValidFile = function(fileName) {
   return fileName && !fileName.startsWith('.');
 }
 exports.isValidFile = isValidFile;
+
+/**
+ * Search for file with the given name, ignoring extensions.
+ * For example, given a desiredFile 'upgrade', will look for
+ * files like upgrade.js and upgrade.sh, and return the filename
+ * of the first found.
```
nanhu95, line 43: Do we think there will ever be a naming conflict issue where someone has a `upgrade.js` intended to be a post upgrade script but a `upgrade.sh` that's meant for something else, and because we return the first found we end up returning the wrong one?

oshi97: hmm yeah that's definitely possible, though it would be pretty weird for somebody to name their files like that. I think eventually we want these filepaths to be configurable through config, so I could just remove this feature, and have jambo only accept upgrade.js and postimport.js for now?

src/utils/fileutils.js
oshi97: I personally prefer this style since it's less nested but don't really mind

src/utils/fileutils.js
nanhu95: I agree, it would definitely be a weird naming scheme. I was mostly asking just to double check the likelihood of something like that happening. I think it's very unlikely that we will run into the scenario we're talking about here in an actual jambo repo, so I think what you've implemented here is okay.

src/utils/fileutils.js
nanhu95: I see what you mean about being less nested. It's definitely a style thing, I'm good with it as is! (My style preference leans towards avoiding if checks where items passing the check aren't actually getting operated on)

nanhu95 (dismissed):
LGTM! I definitely think it's a good idea to rename CustomCommand - I was initially confused by the term.

src/commands/import/themeimporter.js
```diff
@@ -93,27 +92,16 @@ class ThemeImporter{
...
+      const themePath = path.join(this.config.dirs.themes, themeName);
 
       if (addAsSubmodule) {
-        await git.submoduleAdd(themeRepo, localPath);
+        await git.submoduleAdd(themeRepo, themePath);
       } else {
-        await git.clone(themeRepo, localPath);
+        await git.clone(themeRepo, themePath);
       }
+      this._postImport(themePath);
```
cea2aj, line 102: nit: If `themePath` were a property of `ThemeProperty` we could get rid of the param and rename to `_executePostImport`

src/commands/upgrade/themeupgrader.js
```diff
@@ -21,7 +22,7 @@ class ThemeUpgrader {
   constructor(jamboConfig = {}) {
     this.jamboConfig = jamboConfig;
     this._themesDir = jamboConfig.dirs && jamboConfig.dirs.themes;
-    this.upgradeScript = 'upgrade.js';
+    this.postUpgradeFileName = 'upgrade';
```
cea2aj, line 25: Thoughts on using the name 'postupgrade'?

src/commands/import/themeimporter.js
oshi97: I don't think we can set themePath in the constructor because we don't know what themeName is yet

src/commands/upgrade/themeupgrader.js
oshi97: do you mean for the filename or the variable name? if the file name we can't change this or it would break older themes

src/utils/fileutils.js
oshi97: sure that sgtm, updated!

src/commands/import/themeimporter.js
cea2aj: I was thinking we could set as a property outside of the constructor, but I understand if you'd prefer not to do that (I'm not sure if we have a convention about that)

src/commands/upgrade/themeupgrader.js
cea2aj: I was referring to the filename. I wasn't aware of breaking changes, so I'm cool with leaving it

cea2aj (approved):
LGTM

PR #180, opened 2020-12-10
rawconfigvalidator: canonicalize locales when pre-validating

oshi97:
PageConfigsValidator checks that all page configs have
a corresponding locale in the locale_config.
When parsing locales from page config file names, it was
calling parseLocales in i18nutils, which will canonicalize
the locale. However it was no canonicalizing the locale
when pulling them down from the locale_config.json

will cherry-pick to support/v1.9

J=SLAP-936
TEST=manual

test that I can specify a locale like en_gb or en_Gb
on the weight watchers repo and jambo will build the en_GB
page correctly


nanhu95 (approved):
LGTM!

PR #181, opened 2020-12-15
support template data formatter hook

oshi97:
Adds support for a themes/answers-hitchhiker-theme/hooks/templatedata.js,
which is expected to export a default function that defines the custom
data to pass to the templates.

Some yext answers specific details, like injecting the experienceKey into
globalConfig, were being leaked into jambo, probably so that they 
could be part of the template data. In order to not leak these in the
template data formatter hook, these details were removed from our models,
and the logic for them put into jambo's default template data formatter.
The GlobalConfigLocalizer was also removed for this purpose.

J=SLAP-857
TEST=manual,auto

added a custom helper that would fs.writeFile the current template data
into a json file
jambo build-ed a site with and without the templatedata hook,
checked that the 2 json files were identical (eye test + jest)
then checkout current develop version of jambo and test that
the template data without any of these changes was the same as the
other 2

also did the above for yanswers index, index.ja, index.en_GB
also did the above for weight watchers repo when it was still single lang
(no locale_config.json)

tested building a page with universal + vertical standard, and seeing
that the universal page would use the cardType specified in the vertical
page's config


src/commands/build/pagewriter.js
```diff
@@ -83,10 +98,36 @@ module.exports = class PageWriter {
...
+      }
     );
   }
 
+  /**
+   * Returns the data after performing the given template data hook 
+   * transformation on it.
+   *
+   * @param {Object} pageConfig the configuration for the current page
+   * @param {String} relativePath
```
cea2aj, line 110: Can you add a description to `relativePath` to help clarify what its relative to? I'm assuming the standard `relativePath ` from a multilang site dir to the root folder, but I want to confirm

src/commands/build/pagewriter.js
```diff
@@ -83,10 +98,36 @@ module.exports = class PageWriter {
...
+   * @param {Object} params the params object in config for the current locale
+   * @param {Object} globalConfig
+   * @param {Object} pageNameToConfig
```
cea2aj, line 113: Can you add a description here as well?

src/commands/build/pagewriter.js
```diff
@@ -83,10 +98,36 @@ module.exports = class PageWriter {
...
+   * @returns {Object} the additional params to add to the template data
+   */
+  _getTemplateDataFromHook(
+    { pageConfig, relativePath, params, globalConfig, pageNameToConfig }
+  ) {
+    const hookFunction = require(this._templateDataHookPath);
```
cea2aj, line 119: Does this require fail if the file isn't valid JS? If so, do we need to handle that gracefully with an error message?

oshi97: yes we can add an error message!

src/commands/build/pagewriter.js
oshi97: added!

cea2aj (approved):
lgtm!

src/commands/build/pagewriter.js
```diff
@@ -83,10 +99,42 @@ module.exports = class PageWriter {
...
+   * @param {Object} globalConfig
+   * @param {Object} pageNameToConfig
+   * @returns {Object} the additional params to add to the template data
+   */
+  _getTemplateDataFromHook(
+    { pageConfig, relativePath, params, globalConfig, pageNameToConfig }
+  ) {
+    try {
+      const hookFunction = require(this._templateDataHookPath);
+      const hookArgs = {
```
tmeyer2115, line 123: I think it could be good to clean up `hookArgs` a bit. Broadly, we're providing them three types of information:

- Site-level attributes (i.e. `global_config` and `params`)
- Any page-specific configuration (`pageNameToConfig`)
- Information about the current page being considered (`relativePath` and `pageConfig`)

I think it could be good to add some more structure to `hookArgs` along those lines. Potentially we could pass three params:

```
hookFunction(pageMetadata, siteLevelAttributes, pageNameToConfig)
```

where `pageMetadata` contains the page name and `relativePath` and `siteLevelAttributes` contains `global_config` and `params`. Just an idea, but I think right now `hookArgs` feels a bit unstructured and kinda like a random assortment of options.

src/commands/build/pagewriter.js
```diff
@@ -21,6 +21,13 @@ module.exports = class PageWriter {
      * @type {String}
      */
     this._outputDirectory = config.outputDirectory;
+
+    /**
+     * Absolute path to the template data hook, if one is given.
+     * 
+     * @type {String|undefined}
+     */
+    this._templateDataHookPath = config.templateDataHookPath;
```
tmeyer2115, line 30: Nit: The term `templateDataHook` feels a bit generic. What is the hook for? Does it let me read the data or modify it? Alternatively, we could call it something like `templateDataFormatter`.

oshi97: Ah yeah calling it a "hook" isn't really that helpful, I like templateDataFormatter!

src/commands/build/pagewriter.js
oshi97: yeah I like how you organized things! After thinking for a bit we might also want to include things like the current page name, and the locale, in the pageMetadata?

tmeyer2115: That sounds good! If we have the page name and the `pageNameToConfig`, we probably don't need to pass in `pageConfig` any more.

src/commands/build/globalconfiglocalizer.js
```diff
@@ -1,30 +0,0 @@
-const GlobalConfig = require('../../models/globalconfig');
```
oshi97, line 1: Actually on second thought, it would probably be better to keep this class and just use it in PageWriter instead?

src/commands/build/pagesetsbuilder.js
```diff
@@ -54,14 +53,11 @@ module.exports = class PageSetsBuilder {
...
-      const localizedGlobalConfig = new GlobalConfigLocalizer(this._localizationConfig)
-        .localize(this._globalConfig, locale);
-
       pageSets.push(new PageSet({
         locale: locale,
         pages: this._buildPages(pageConfigs, templates),
-        globalConfig: localizedGlobalConfig,
-        params: this._localizationConfig.getParams(locale)
+        globalConfig: this._globalConfig,
+        localizationConfig: this._localizationConfig
```
tmeyer2115, line 60: Just confirming, this is the JSON.parse'd locale_config?

src/commands/build/pagewriter.js
```diff
@@ -37,6 +45,13 @@ module.exports = class PageWriter {
       : '';
     console.log(`Writing files${localeMessage}`);
 
+    const pageSetData = {
+      localizationConfig: pageSet.getLocalizationConfig(),
+      globalConfig: pageSet.getGlobalConfig().getConfig(),
```
tmeyer2115, line 50: Is `pageSet.getGlobalConfig().getConfig()` a new addition or does the GlobalConfig already have a method called `getConfig`? Seems kinda redundant and confusing.

src/commands/build/pagewriter.js
```diff
@@ -63,30 +77,115 @@ module.exports = class PageWriter {
...
+  _getTemplateDataFromFormatter({
+    relativePath,
+    locale,
+    pageName,
+    globalConfig,
+    currentLocaleConfig,
+    pageNameToConfig
+  }) {
+    try {
+      const hookFunction = require(this._templateDataFormatter);
```
tmeyer2115, line 170: `formatterFunction`

src/commands/build/pagewriter.js
```diff
@@ -63,30 +77,115 @@ module.exports = class PageWriter {
...
+      return this._getTemplateDataFromFormatter({
+        relativePath,
+        currentLocaleConfig,
+        globalConfig,
+        pageNameToConfig,
+        locale,
+        pageName
+      });
+    }
+    const localizedGlobalConfig =
```
tmeyer2115, line 112: The `PageWriter` class is doing A LOT, even before your change. Could we make a default `templateDataFormatter` and move it to it's own file? We can then do some defaulting above: 
```
this._templateDataFormatter = config.templateDataFormatter || whateverTheDefaultIsCalled;
```

Having the default formatter separated out will allow for more sane unit testing too.

src/commands/build/sitesgenerator.js
```diff
@@ -175,9 +175,12 @@ class SitesGenerator {
         page.setTemplateContents(processedTemplate);
       }
 
+      const templateDataFormatter =
+        path.resolve(config.dirs.themes, config.defaultTheme, 'hooks', 'templatedata.js');
```
tmeyer2115, line 179: What do you think about calling it `templatedataformatter.js` or is that too long winded? We want the name to make it clear it's formatting HBS data for the templates.

src/models/pageset.js
```diff
@@ -65,12 +66,12 @@ module.exports = class PageSet {
...
   /**
-   * Returns the params
+   * Returns the localization config
    *
-   * @returns {Object} params
+   * @returns {LocalizationConfig}
    */
-  getParams() {
-    return this.params;
+  getLocalizationConfig() {
```
tmeyer2115, line 73: Should a `PageSet` need access to the entire `locale_config` or just the bit that pertains to it's `locale`?

tmeyer2115 (commented):
Looking through all this code again, I think we pretty desperately need TypeScript in Jambo. It's really hard to figure out how all these classes relate to one another. I see a lot of params listed as `Object` too, which makes the code hard to follow. The intermixing of classes and raw Objects seems no bueno. 

I think we need some strongly typed domain models that correspond to the problem space. None of that rant is related to this PR though, haha. cc: @bryanreed

src/commands/build/pagesetsbuilder.js
oshi97: nope this is the LocalizationConfig model!

src/commands/build/pagewriter.js
oshi97: this is from before! getConfig() gets the raw object in the GlobalConfig model

src/commands/build/sitesgenerator.js
oshi97: hmm it felt a little long to me before but I don't think it's the worst, will update!

src/models/pageset.js
oshi97: we can definitely restrict it to just that!

src/commands/build/pagewriter.js
tmeyer2115: gotcha! I think that's somewhat confusing, so maybe we can address as part of the TypeScript refactor i'm hoping for.

src/commands/build/pagewriter.js
oshi97: yeah I think that makes a lot of sense! before the one PageWriter unit test was actually only testing generating the template data which was a little weird

src/commands/build/templateargsformatter.js
```diff
@@ -0,0 +1,79 @@
+const defaultTemplateDataFormatter = require('./defaulttemplatedataformatter');
+const fs = require('fs');
+
+/**
+ * TemplateArgsFormatter is responsible for building the arguments that are
+ * sent to the handlebars templates.
+ */
+module.exports = class TemplateArgsFormatter {
```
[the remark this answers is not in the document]
oshi97: does TemplateArgsBuilder work?

src/commands/build/defaulttemplatedataformatter.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * This is the default data formatting for the data passed into the
+ * handlebars templates Jambo builds.
+ *
+ * @param {Object} pageMetadata contents below
```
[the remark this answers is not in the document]
oshi97: the line length made this a little hard but I can adjust

tmeyer2115 (approved):

PR #182, opened 2020-12-17
themeimporter: fix not being able to init with theme

nanhu95:
Fixes the issue where a user cannot run a `jambo init` with the
theme specified.

The usage of the theme importer as a constructor was originally
incorrect in the repository scaffolder; the change is made here
to use it correctly.

TEST=manual
Created a new directory locally, `cd`-ed into it, and ran
`jambo init --theme answers-hitchhiker-theme`. Saw that the
jambo repo was initialized correctly and that the theme was
also imported into the repo.

### DIFF
--- a/src/commands/import/themeimporter.js
+++ b/src/commands/import/themeimporter.js
@@ -73,7 +73,7 @@ class ThemeImporter{
   }
 
   execute(args) {
-    this._import(args.theme, args.addAsSubmodule)
+    this.import(args.theme, args.addAsSubmodule)
       .then(console.log);
   }
 
--- a/src/commands/import/themeimporter.js
+++ b/src/commands/import/themeimporter.js
@@ -87,7 +87,7 @@ class ThemeImporter{
    *                            containing the new submodule's local path. If the addition
    *                            failed, a Promise containing the error.
    */
-  async _import(themeName, addAsSubmodule) {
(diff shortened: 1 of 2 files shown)


src/commands/import/themeimporter.js
```diff
@@ -87,7 +87,7 @@ class ThemeImporter{
    *                            containing the new submodule's local path. If the addition
    *                            failed, a Promise containing the error.
    */
-  async _import(themeName, addAsSubmodule) {
+  async import(themeName, addAsSubmodule) {
```
tmeyer2115, line 90: Will renaming this break the `import` command? It looks like the `execute` method expects a `this._import`.

tmeyer2115 (commented):
We should make a hotfix branch for Jambo. That branch will be the target of this PR. Under gitflow, we don't commit directly to master unless it's a new release.

src/commands/import/themeimporter.js
nanhu95: Good catch! I thought using `F2` to rename would have caught the usage in `exec`.

nanhu95:
> We should make a hotfix branch for Jambo. That branch will be the target of this PR. Under gitflow, we don't commit directly to master unless it's a new release.

Done!

tmeyer2115 (approved):

PR #183, opened 2020-12-17
Upgrade command's `describe` should list a default for `branch`.

tmeyer2115:
The `upgrade` command implicitly assumes a default of `master` for the
`branch` parameter. That was not getting communicated to SGS because we left it
out of the command's `describe`. I added some additional logic so that we don't
list any default for `branch` when the repo has no Theme yet. It looked a bit
weird to have no options listed for `branch`, but a default of `master`.

J=SLAP-978
TEST=manual

Ran `describe` on an empty repo and saw no default listed for `branch`. Ran
`describe` on a repo with a Theme and saw a default of `master` reported. Also
ensured that running `upgrade` with no `branch` resulted in `master` being
used.


nanhu95 (approved):
LGTM!

PR #184, opened 2020-12-17
Version 1.9.8.

tmeyer2115:
- Fixed a bug that caused `init` to fail when the `--theme` flag was used.
(#182)
- Update the `describe` command so that it correctly lists `master` as the
default of `upgrade.params.branch`. (#183)


oshi97 (approved):
lgtm!

PR #185, opened 2020-12-18
Add legacy support for the older style of custom command importing.

tmeyer2115:
Jambo should support both the current and old style of importing a custom
command. The legacy style, which can be seen in Theme v1.16, didn't export a
class, but a function. The function took in a jambo config and provided an
instance of the custom command's class. Further, back then, the command class
did not have any static members.

The `CommandImporter` is now able to detect legacy command imports. It wraps
them in an implementation of the latest `Command` interface. Note that all
legacy imports were characterized by a function starting with 'jamboConfig =>'.

J=SLAP-942
TEST=manual

Tested the following for sites on Theme v1.16 and Theme v1.18:
- `jambo --help` produced the expected output.
- The `--help` flag on the `card` and `directanswercard` commands worked as
  expected.
- The `describe` command had the correct output for `card` and
  `directanswercard`.
- The `directansercard` command worked as expected.
- The `card` command worked as expected.

### DIFF
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -108,7 +108,7 @@ class CommandImporter {
    * @return {boolean} Boolean indicating if this is a legacy command import.
    */
   _isLegacyImport(requiredModule) {
-    return requiredModule.toString().startsWith('jamboConfig =>');
+    return !('prototype' in requiredModule);
   }
 
   /**
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -30,7 +30,12 @@ class CommandImporter {
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
         .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
         .forEach(filePath => {
-          const commandClass = require(filePath);
+          const requiredModule = require(filePath);
+
+          const commandClass = this._isLegacyImport(requiredModule) ?
+            this._handleLegacyImport(requiredModule) :
+            requiredModule;
+
           if (this._validateCustomCommand(commandClass)) {
             customCommands.push(commandClass);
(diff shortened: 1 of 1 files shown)


nanhu95 (approved):
LGTM! Just one very small nit: in the jsdocs, `require'd` => `required`

tmeyer2115:
> LGTM! Just one very small nit: in the jsdocs, `require'd` => `required`

Ah, having it as `require'd` was intentional! I've seen that in other JS articles/docs to indicate something that was pulled in via `require`.

nanhu95:
> > LGTM! Just one very small nit: in the jsdocs, `require'd` => `required`
> 
> Ah, having it as `require'd` was intentional! I've seen that in other JS articles/docs to indicate something that was pulled in via `require`.

Oh gotcha! Nice to know. Then everything LGTM!

src/commands/commandimporter.js
```diff
@@ -95,5 +100,50 @@ class CommandImporter {
...
   }
+  
+  /**
+   * Verifies if the require'd module corresponds to a legacy command import.
+   * 
+   * @param {any} requiredModule The require'd module.
+   * @return {boolean} Boolean indicating if this is a legacy command import.
+   */
+  _isLegacyImport(requiredModule) {
+    return requiredModule.toString().startsWith('jamboConfig =>');
```
oshi97, line 111: I was wondering if there might be a more javascript-y way to do this? I did some searching and maybe we could check for requiredModule.prototype (from [here](https://stackoverflow.com/questions/40922531/how-to-check-if-a-javascript-function-is-a-constructor)) 
I'm fine with this implementation though!

tmeyer2115: Updated!

oshi97 (approved):

PR #186, opened 2020-12-29
Ignore node modules partials

cea2aj:
Ignore node_modules when registering partials

When adding the [theme testing site](https://github.com/yext/answers-hitchhiker-theme/pull/555), I ran into a problem as jambo would try to register the files inside node modules. Ignoring paths that contain node modules when registering partials fixes this issue.

J=none
TEST=manual

Build the testing repo and observe a successful build

### DIFF
--- a/src/models/partialsregistry.js
+++ b/src/models/partialsregistry.js
@@ -1,6 +1,6 @@
 const fs = require('file-system');
 
-const { isValidFile } = require('../utils/fileutils');
+const { isValidFile, isValidPartialPath } = require('../utils/fileutils');
 const Partial = require('./partial');
 
 /**
@@ -44,7 +44,7 @@ module.exports = class PartialsRegistry {
   }
 
   /**
-   * Builds all partials in the provided path. If the path is a directory,
+   * Builds all valid partials in the provided path. If the path is a directory,
    * the useFullyQualifiedName parameter dictates if the path's root will be
    * included in the partial naming scheme.
    *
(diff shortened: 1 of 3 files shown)


src/models/partialsregistry.js
```diff
@@ -58,7 +58,7 @@ module.exports = class PartialsRegistry {
     const pathExists = fs.existsSync(partialsPath);
     if (pathExists && !fs.lstatSync(partialsPath).isFile()) {
       fs.recurseSync(partialsPath, (path, relative, filename) => {
-        if (isValidFile(filename)) {
+        if (isValidFile(filename) && !path.includes('node_modules')) {
```
oshi97, line 61: I think this would break if I want to register a path called snode_modulest? you could use startsWith instead of includes but that wouldn't work for things like test-site/node_modules, if that's something we need to worry about. One solution I thought of was prefixing path with `/` and checking for `/node_modules/` but that's a little weird

cea2aj: I think ignoring `/node_modules/` actually makes sense. The recurse sync supports filtering on the files to recurse, but I couldn't for the life of me get it to ignore node modules (We need to ignore a top-level node modules and the node-modules in the sample-site)

oshi97: okay sure that sounds fine to me then! we'll probably want to add some kind of unit test for this even if it's small

tests/utils/fileutils.js
```diff
@@ -54,6 +55,20 @@ describe('isValidFile properly determines if files are valid', () => {
   });
 });
 
+describe('isValidPartialPath properly determines if paths are valid', () => {
```
oshi97, line 58: kind of a nit but could we add a test case for when the path starts with node_modules?

cea2aj: sure

oshi97 (approved):
lgtm!

PR #190, opened 2021-01-29
Remove ability of `upgrade` command to fetch Theme branches.

tmeyer2115:
The `ThemeUpgrader.describe` method would attempt to fetch the branches for the
default Theme. The branches would be listed as the various options for the
`--branch` flag. The problem was that the method for fetching the branches
did not work unless the Theme was imported as a sub-module. Rather than fix
the logic to work for all cases, we decided this was outside Jambo's realm of
responsibility. If someone is using a Theme, they should be able to tell Jambo
which branch to use. Jambo does not need to supply them with a list of the
valid options. As such, `describe` now lists the `branch` param as a normal
string, not asingleoption.

This is considered a bug fix, so the Jambo version was incremented by x.x.1.

J=SLAP-1047
TEST=auto,manual

Reran all Jambo unit tests. Verified the following on a sample site:
- The `describe` command has the expected output for `upgrade`.
- The `upgrade --help` command produces the expected output.
- Running `upgrade` without `--branch` pulls the latest commit off master.
- Running `upgrade` with a `--branch` pulls the latest commit off the branch.

### DIFF
--- a/src/commands/upgrade/themeupgrader.js
+++ b/src/commands/upgrade/themeupgrader.js
@@ -55,17 +55,6 @@ class ThemeUpgrader {
   }
 
   static async describe(jamboConfig) {
-    const branches = await this._getThemeBranches(jamboConfig);
-
-    const branchParam = {
-      displayName: 'Branch of theme to upgrade to',
-      type: 'singleoption',
-      options: branches
-    }
-    if (branches.length) {
-      branchParam.default = 'master';
-    }
-
     return {
       displayName: 'Upgrade Theme',
       params: {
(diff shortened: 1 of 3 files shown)


src/commands/upgrade/themeupgrader.js
```diff
@@ -55,17 +55,6 @@ class ThemeUpgrader {
   }
 
   static async describe(jamboConfig) {
-    const branches = await this._getThemeBranches(jamboConfig);
```
oshi97, line 58: I think we can make this method synchronous again? Now that we don't have to fetch git branches

tmeyer2115: Ah, I kept this as `async` because the `Command` class itself uses `async` in the `describe` signature.

cea2aj (approved):
lgtm

PR #195, opened 2021-02-26
Update getRepoForTheme to use HTTPS instead of SSH.

tmeyer2115:
This PR updates the getRepoForTheme method to use the HTTPS protocol instead
of SSH. Recently, the Theme was made public. The hope was that this meant
the Preview containers in CBD would not need an SSH key anymore. But, if the
SSH protocol is used, an SSH key is still required, even though the Theme
repo is now public.

TEST=manual

Removed my local machine's SSH key from Github and verified that I could still
import and upgrade the Theme in a Jambo site.


bjulian5 (approved):
lgtm

PR #197, opened 2021-03-01
v1.10.2

tmeyer2115:
### Bug Fixes
- The HTTPS protocol is now used when cloning the Hitchhiker Theme. The Theme was recenty open sourced, so there will be no credential issues. (#195)
- The `describe` command no longer lists incorrect `path` options for the `override` command. Previously, the options had an incorrect root prefix. (#196)


oshi97 (approved):
💯

PR #198, opened 2021-04-07
Only load javascript files for custom commands

cea2aj:
Ensure that files end with '.js' before trying to load the module

This fixes a bug where jambo would crash if a .DS_Store file was present in the commands folder.

J=none
TEST=manual

Test that jambo no longer crashes if a .DS_Store file is present in the command folder. Test the `jambo --help` command, the `jambo describe` command, and the custom vertical command from the theme.

### DIFF
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -29,7 +29,6 @@ class CommandImporter {
       fs.readdirSync(mergedDirectory)
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
         .filter(directoryPath => directoryPath.endsWith('.js'))
-        .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
         .forEach(filePath => {
           const requiredModule = require(filePath);
 
--- a/src/commands/commandimporter.js
+++ b/src/commands/commandimporter.js
@@ -28,6 +28,7 @@ class CommandImporter {
       const mergedDirectory = this._mergeCommandDirectories(commandDirectories);
       fs.readdirSync(mergedDirectory)
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
+        .filter(directoryPath => directoryPath.endsWith('.js'))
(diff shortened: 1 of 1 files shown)


oshi97 (approved):
nice!

src/commands/commandimporter.js
```diff
@@ -28,6 +28,7 @@ class CommandImporter {
       const mergedDirectory = this._mergeCommandDirectories(commandDirectories);
       fs.readdirSync(mergedDirectory)
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
+        .filter(directoryPath => directoryPath.endsWith('.js'))
         .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
```
tmeyer2115, line 32: If we're checking that the path ends with .js, do we need the `isFile` check? Seems like it may be a bit redundant.

cea2aj: The file check is probably unnecessary. I just updated the PR to remove it

oshi97 (approved):

src/commands/commandimporter.js
```diff
@@ -28,7 +28,7 @@ class CommandImporter {
       const mergedDirectory = this._mergeCommandDirectories(commandDirectories);
       fs.readdirSync(mergedDirectory)
         .map(directoryPath => path.resolve(mergedDirectory, directoryPath))
-        .filter(directoryPath => fs.lstatSync(directoryPath).isFile())
+        .filter(directoryPath => directoryPath.endsWith('.js'))
```
oshi97, line 31: sorry just had a thought, will this error out if somebody names their folder "myFolder.js"?

oshi97: or rather makes a folder in their custom commands folder that ends in ".js"

cea2aj: ya, it does. I'll add the file check back

oshi97 (approved):

PR #201, opened 2021-04-13
use new HbsHelperParser to parse out translation helpers

oshi97:
This commit adds HbsHelperParser, and uses it to parse out
translation helpers instead of regex. Under the hood, this uses
Handlebars.Visitor, which accepts a Handlebars AST generated by
Handlebars.parse().

J=SLAP-1178
TEST=manual,auto

update tests to have multi line translation helpers
hbs comments that contain translation helpers are now also
correctly ignored! html comments are not ignored (as it should be)

test jambo building a site with multi line translation helpers
see that using an older version of jambo would not work with multi line helpers,
but adding in this change would fix that


src/handlebars/hbshelperparser.js
```diff
@@ -0,0 +1,104 @@
...
+ * requested hbs helpers, and parses out an array of all
+ * hbs helpers found of the requested types.
+ */
+class HbsHelperParser {
+  constructor(requestedHelpers) {
+    /**
+     * The helpers that should be parsed out.
+     * @type {string[]}
+     */
+    this.requestedHelpers = requestedHelpers;
```
cea2aj, line 14: nit: what do you think of renaming to 'helpersToParse'?

src/handlebars/hbshelperparser.js
```diff
@@ -0,0 +1,104 @@
...
+    this.vistor = new Handlebars.Visitor();
+
+    /**
+     * _handleMustacheStatement() is dispatched on all MustacheStatement nodes.
+     * @type {Function}
+     */
+    this.vistor.MustacheStatement = this._handleMustacheStatement.bind(this);
+
+    /**
+     * An accumulator for helper statements found in a template.
```
cea2aj, line 28: nit: What do you think of shortening to 'Helper statements found in a template'? The word 'accumulator' makes me think that it's related to `Array.prototype.reduce()` which isn't the case

src/handlebars/hbshelperparser.js
```diff
@@ -0,0 +1,104 @@
...
+     * @type {hbs.AST.MustacheStatement[]}
+     */
+    this.helperStatements = [];
+  }
+
+  /**
+   * Parses all translate invocation strings from a handlebars template.
```
cea2aj, line 35: nit: it's not necessarily parsing translate invocations

src/handlebars/hbshelperparser.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * Gets the original value for a given MustacheStatement.
+   * 
+   * @param {hbs.AST.MustacheStatement} statement the statement to get the original of
+   * @param {number[]} lineEndIndices the line end indices of the original template
+   * @param {string} template the original template
+   */
+  _getOriginalValue(statement, lineEndIndices, template) {
```
cea2aj, line 92: Would it be more accurate for this to be named `_getOriginalHelperCall`?

src/handlebars/hbshelperparser.js
oshi97: shoot missed this when renaming stuff (originally this class was InvocationExtractor)

src/handlebars/hbshelperparser.js
oshi97: yes that's more specific!

src/handlebars/hbshelperparser.js
oshi97: yeah I like that!

cea2aj (approved):
LGTM!

PR #207, opened 2021-04-15
rename addAsSubmodules to useSubmodules, false by default

oshi97:
This commit renames the --addAsSubmodules param of jambo import and init to
--useSubmodules, and changse the default to false.

J=SLAP-1135
TEST=manual

test that I can import a theme with useSubmodules, and it will be a submodule
test jambo import --theme answers-hitchihker-theme and see it will be imported
as regular checked in files, with the a .git folder removed. check that I can commit
the theme changes without any errors/warning messages, which you would see if
you tried the .git folder wasn't removed

test importing a theme on init, but as submodule and as regular files

### DIFF
--- a/src/commands/import/themeimporter.js
+++ b/src/commands/import/themeimporter.js
@@ -11,6 +11,7 @@ const { ArgumentMetadata, ArgumentType } = require('../../models/commands/argume
 const { CustomCommand } = require('../../utils/customcommands/command');
 const { CustomCommandExecuter } = require('../../utils/customcommands/commandexecuter');
 const { searchDirectoryIgnoringExtensions } = require('../../utils/fileutils');
+const fsExtra = require('fs-extra');
 
 /**
  * ThemeImporter imports a specified theme into the themes directory.
@@ -41,10 +42,9 @@ class ThemeImporter {
         description: '(deprecated: specify the themeUrl instead)'
           + ' the name of the theme to import',
       }),
-      addAsSubmodule: new ArgumentMetadata({
+      useSubmodules: new ArgumentMetadata({
         type: ArgumentType.BOOLEAN,
-        description: 'import the theme as a submodule',
-        defaultValue: true
+        description: 'import the theme as a submodule'
       }),
     }
   }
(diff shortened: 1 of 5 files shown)


src/commands/import/themeimporter.js
```diff
@@ -41,10 +42,9 @@ class ThemeImporter {
         description: '(deprecated: specify the themeUrl instead)'
           + ' the name of the theme to import',
       }),
-      addAsSubmodule: new ArgumentMetadata({
+      useSubmodules: new ArgumentMetadata({
         type: ArgumentType.BOOLEAN,
-        description: 'import the theme as a submodule',
-        defaultValue: true
+        description: 'import the theme as a submodule'
```
cea2aj, line 47: Is there a reason for not setting the default here and in the describe command as false?

src/commands/init/initcommand.js
```diff
@@ -26,7 +26,7 @@ class InitCommand {
           + ' the name of a theme to import during the init',
         isRequired: false
       }),
-      addThemeAsSubmodule: new ArgumentMetadata({
+      useSubmodules: new ArgumentMetadata({
         type: ArgumentType.BOOLEAN, 
         description: 'if starter theme should be imported as submodule',
         defaultValue: true
```
cea2aj, line 32: defaultValue should be false

src/commands/init/initcommand.js
```diff
@@ -48,8 +48,8 @@ class InitCommand {
           type: 'singleoption',
           options: importableThemes
         },
-        addThemeAsSubmodule: {
-          displayName: 'Add Theme as Submodule',
+        useSubmodules: {
+          displayName: 'Use Submodules',
           type: 'boolean',
           default: true
```
cea2aj, line 54: I believe this should be `default: false` as well

src/commands/import/themeimporter.js
oshi97: it looks like if we want a param to be not used by default, we've just not been including the defaultValue. Right now there's no difference between `jambo import --useSubmodules false ...` and `jambo import ...`, both functionally, and from a cbd ui perspective, but we might want to keep this distinction?

src/commands/init/initcommand.js
oshi97: thanks

cea2aj (approved):
lgtm

PR #210, opened 2021-04-28
factor out buildJamboCLI() from cli.js

oshi97:
This commit factors out the buildJamboCLI() method
from cli.js. No functional changes have been made.
This method will be used by the upcoming acceptance test
infrastructure, to create test jambo instances.

J=SLAP-992
TEST=manual

test that I can still build pages with jambo
test that jambo still logs 'You must provide Jambo with a command.'
if you don't enter a command
test that custom commands still work

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,13 +1,6 @@
 #!/usr/bin/env node
-
-const fs = require('file-system');
-const path = require('path');
-
-const { parseJamboConfig } = require('./utils/jamboconfigutils');
 const { exitWithError } = require('./utils/errorutils');
-const CommandRegistry = require('./commands/commandregistry');
-const YargsFactory = require('./yargsfactory');
-const CommandImporter = require('./commands/commandimporter');
+const buildJamboCLI = require('./buildJamboCLI');
 
 // Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -17,67 +10,5 @@ process.on('uncaughtException', err => {
...
-      jamboConfig.dirs.output,
-      path.join(jamboConfig.dirs.themes, jamboConfig.defaultTheme)) :
-    new CommandImporter(jamboConfig.dirs.output);
-
-  commandImporter.import().forEach(customCommand => {
-    commandRegistry.addCommand(customCommand)
-  });
-}
+const jambo = buildJamboCLI(process.argv)
+jambo && jambo.argv;
```
cea2aj, line 14: nit: It's not very clear to me what `jambo.argv` is actually doing. According to the docs, it looks like `.parse()` should work instead. What you do you think about using that?

Maybe something like: `const jamboCLI = buildJamboCLI(process.argv);
jamboCLI && jamboCLI.parse();`

oshi97: I like it, updated! I also asked Tom something similar a while back https://github.com/yext/jambo/pull/130#discussion_r499838521

cea2aj (commented):
lgtm

cea2aj (approved):

PR #212, opened 2021-04-29
Add test-themes folder and "basic" theme

oshi97:
This commit adds the first test theme to Jambo.
This will be used by an upcoming acceptance test.
It has close to the minimal number of files needed for the
[init -> import -> page -> build] flow to work.

It also adds the setup and cleanup functions for the test-themes folder.
These functions will be used by the upcoming acceptance test infrastructure.

J=SLAP-992
TEST=auto

ran acceptance tests using this test theme, the tests will be added in a future commit


package.json
```diff
@@ -45,7 +45,8 @@
   "jest": {
     "verbose": true,
     "testMatch": [
-      "**/tests/**/*.js"
+      "**/tests/**/*.js",
+      "!**/tests/acceptance/**"
```
cea2aj, line 49: nit: is there a reason why this isn't in "testPathIgnorePatterns"?

oshi97: nope! updated

tests/acceptance/setup/playground.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Transform all theme folders under test-themes back into regular folders,
+ * so they can be tracked properly by jambo's git repo.
+ */
+function cleanupTestThemes() {
+  const testThemesDir = path.resolve(__dirname, '../test-themes');
+  const testThemes = fs.readdirSync(testThemesDir);
+  testThemes.forEach(themeName =>{
```
cea2aj, line 33: super nit: I'm used to seeing a space after the arrow `=>`

oshi97: no that's a good catch! I added an eslint rule since this was the only instance of the inconsistent arrow spacing

tests/acceptance/test-themes/basic/templates/universal-standard/page-config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "pageTitle": "universal page",
+  "componentSettings": {
+    "DirectAnswer": {
+      "dummyConfig": "dummy config"
+    }
+    /**
+    "QASubmission": {
+      "entityId": "", // Set the ID of the entity to use for Q&A submissions, must be of entity type "Organization"
```
cea2aj, line 9: While we are here, can we add JSON with comments support like in https://github.com/yext/answers-hitchhiker-theme/pull/689?

tests/acceptance/test-themes/basic/postimport.js
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Updates the defaultTheme field in the jambo.json.
+ * 
+ * @param {Object} jamboConfig 
+ * @param {string} themeName 
+ */
+function updateDefaultTheme(jamboConfig, themeName) {
+  if (jamboConfig.defaultTheme !== themeName) {
+    const updatedConfig = assign({ defaultTheme: themeName }, jamboConfig);
+    fs.writeFileSync('jambo.json', stringify(updatedConfig, null, 2));
```
cea2aj, line 16: Why are we writing to 'jambo.json', but the jamboConfig that this is based on is from '../../jambo.json'? Are we not reading and writing to the same file?

oshi97: The require is relative to the filepath of the postimport script, `playground/themes/basic/postimport.js`, but fs.writeFileSync is relative to the current working directory, `playground/`

tests/acceptance/test-themes/basic/templates/universal-standard/page-config.json
oshi97: yes definitely!

oshi97: I tried adding the .gitattributes file, then even adding a commit where I deleted page-config.json and another one where I readded it, but the json-with-comments support doesn't seem to be working?

cea2aj: looks like you fixed it!

oshi97: github caches the highlighting, fixed now!

tests/acceptance/test-themes/basic/postimport.js
cea2aj: Do you think it would make sense for this function to take in a path to the jambo config to edit? That way it would handle the read, modification, and the write?

oshi97: It feels a little weird to read in the jambo config in two places

oshi97: Sure, the path is always "jambo.json" though so I don't think that needs to be a parameter

cea2aj: that works!

cea2aj (approved):

PR #213, opened 2021-04-29
Add "basic-flow" acceptance suite and testing infra

oshi97:
This commit adds the basic-flow acceptance test suite,
and the acceptance testing infrastructure it needs.

This commit removes testing for node 10.x, and adds in testing for 16.x
jambo was already incompatible with node 10.x, our unit tests just weren't covering incompatible
portions of the code. The basic-flow acceptance test now blanket covers some areas of code
that are incompatible with 10.x.

J=SLAP-992
TEST=auto


.github/workflows/run-tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
```
cea2aj, line 15: Just a heads up, the theme doesn't support node 16. We use an older version of node sass which doesn't support it. That shouldn't affect Jambo though

.github/workflows/run-tests.yml
```diff
@@ -21,4 +21,6 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm ci
+    - run: git config --global user.email "slapshot@yext.com"
+    - run: git config --global user.name "Jambo run-tests.yml"
```
cea2aj, line 25: nit: Is there a reason for not using "Team Slapshot" as the name?

package.json
```diff
@@ -35,21 +35,29 @@
...
   },
   "jest": {
     "verbose": true,
+    "setupFilesAfterEnv": [
+      "./tests/setup/setup.js"
+    ],
     "testMatch": [
-      "**/tests/**/*.js"
+      "**/tests/**/*.js",
+      "!**/tests/acceptance/**/*.js",
```
cea2aj, line 55: nit: can we put the paths to ignore in `testPathsIgnorePatters'

oshi97: We can't because putting tests/acceptance there will ignore the entire path, even if we include tests/acceptance/suites in testMatch

.github/workflows/run-tests.yml
oshi97: This isn't currently possible, but say we push to a repo accidentally - this is a little easier to trace than just "Team Slapshot"

oshi97: also we someday might now own jambo (though we'll have to update other parts of the code that reference us then anyways)

.github/workflows/run-tests.yml
oshi97: thanks for the heads up!

tests/acceptance/setup/playground.js
```diff
@@ -34,4 +36,32 @@ function cleanupTestThemes() {
...
+    fsExtra.removeSync(playgroundDir);
+  }
+
+  try {
+    fsExtra.mkdirpSync(playgroundDir);
+    await setupTestThemes();
+    chdir(playgroundDir);
+    await testFunction(new TestInstance());
+  } catch (err) {
+    cleanup();
```
cea2aj, line 63: nit: can we put the `cleanup()` inside a finally block?

oshi97: yes that's much better! I didn't know these existed

.github/workflows/run-tests.yml
cea2aj: It's not possible? Isn't just setting a string?

cea2aj: And do you mean "might not own"?

package.json
cea2aj: ah, ok

tests/acceptance/setup/playground.js
cea2aj: ya, I think a cleanup function is a standard use case of "finally'

.github/workflows/run-tests.yml
oshi97: sorry my wording was poor, I mean't it's not currently possible for our run-test workflow to push to a repo accidentally.

tests/acceptance/suites/basic-flow.js
```diff
@@ -0,0 +1,27 @@
...
+// silence jambo's noisy output
+console.log = jest.fn();
+
+it('can init -> import -> create page -> build', () => runInPlayground(async t => {
+  await t.jambo('init');
+  await t.jambo('import --themeUrl ../test-themes/basic');
+  await t.jambo('page --name index --template universal-standard');
+  await t.jambo('build');
+  const indexPage = fs.readFileSync('public/index.html', 'utf-8');
+  const expectedPage =
```
cea2aj, line 15: What are your thoughts on moving this to a fixture?

oshi97: I thought about it but... I was a little worried about having way too many fixtures? I was originally imagining having a few tests in this basic-flow.js file, and then having several fixtures for it in an acceptance/fixtures folder could quickly grow out of control. Now that I think about it more, though, this is easily fixed by just making an acceptance/fixtures/basic-flow folder.

cea2aj (approved):
🍓 🍓 🍓
Cool solution!

PR #216, opened 2021-05-07
add Coveralls github workflow

oshi97:
Adds a code coverage workflow using Coveralls.
Because the acceptance tests are ran using jest, and are run
in the main thread (i.e. no spawnSyncs), the acceptance tests
count towards our coverage. Acceptance test coverage is probably
of a lower overall "quality" than our unit test coverage, since it's
not as rigid about checking output values.

Also adds a code coverage badge to the readme.

J=SLAP-873
TEST=auto

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,30 @@
+# This workflow will run our tests, generate an lcov code coverage file,
+# and send that coverage to Coveralls 
+
+name: Code Coverage
+
+on:
+  push:
+    branches-ignore: dev/*
+  pull_request:
+
+jobs:
+  Coveralls:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16.x]
+
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: git config --global user.email "slapshot@yext.com"
+    - run: git config --global user.name "Jambo run-tests.yml"
+    - run: npx jest --coverage
```
oshi97, line 28: if we want we can only collect coverage from unit tests, what do y'all think?

tmeyer2115: How well does Coveralls gauge coverage from acceptance tests? It seems like that could be a difficult thing to get right. My initial reaction is that if Coveralls doesn't do well with acceptance tests, we should stick to just the unit tests.

tmeyer2115 (approved):

PR #224, opened 2021-05-19
Do not run the `babel` helper in Development Mode.

tmeyer2115:
This PR adds a check to the `babel` helper to see if we are currently in the Development Mode. If
so, the enclosed source code is not run through Babel, it is returned verbatim. This is a preview
optimization, since the Babel-ification is a relatively slow operation. It's not necessary for
preview either, assuming devs are not previewing their changes in IE11.

J=SLAP-1312
TEST=manual

Ensured that when in Development mode, the helper was a simple pass-through. In Production mode,
the enclosed code was Babel-ified.

### DIFF
--- a/src/handlebars/registerhbshelpers.js
+++ b/src/handlebars/registerhbshelpers.js
@@ -83,7 +83,7 @@ module.exports = function registerHbsHelpers(hbs) {
     if (process.env.IS_DEVELOPMENT_PREVIEW === 'true' ) {
       return srcCode;
     } else {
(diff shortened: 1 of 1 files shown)


cea2aj:
If someone does want to test the page in IE11, then they will need to publish to staging, correct?

cea2aj (approved):

tmeyer2115:
> If someone does want to test the page in IE11, then they will need to publish to staging, correct?

Correctomundo

PR #226, opened 2021-05-27
Remove outdated `sample_site` from Jambo.

tmeyer2115:
This PR removes the sample site that was included with Jambo. This sample is very much out of date. It was
created at Jambo's inception. The Hitchhiker Training and Theme provide much better guidance on how to
use Jambo.

J=SLAP-1239
TEST=none


cea2aj (approved):
lg

PR #228, opened 2021-06-03
use npmlog for logging with colors

oshi97:
This commit refactors jambo's logging to use npmlog,
which is the logger used by npm (surprise!).
It adds a magenta 'jambo' log heading to differentiate it from
npm, which uses a white on black 'npm' heading.

J=SLAP-226
TEST=manual

test a regular jambo build
test jambo build when it throws an error because there's no jambo.json
test jambo page that is missing required arguments
test the custom card command in the theme
see that the tested logs look as expected, and contain the same information as before
jambo help prints normally
jambo version prints normally
warning is given if custom command is malformatted

test that you can set the npmlog log level anywhere, and that log level will be reflected globally
in the future we can let jambo set different log levels or even silence logging

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -2,8 +2,9 @@
 const { exitWithError } = require('./utils/errorutils');
 const buildJamboCLI = require('./buildJamboCLI');
 
+// Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
+process.on('unhandledRejection', exitWithError);
+process.on('uncaughtException', exitWithError);
+
 const jambo = buildJamboCLI(process.argv);
-jambo && jambo.parseAsync().catch(err => {
-  // Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
-  exitWithError(err);
-});
+jambo && jambo.parseAsync().catch(exitWithError);
--- a/src/cli.js
+++ b/src/cli.js
@@ -2,13 +2,8 @@
 const { exitWithError } = require('./utils/errorutils');
 const buildJamboCLI = require('./buildJamboCLI');
 
-// Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
-process.on('unhandledRejection', err => {
-  exitWithError(err);
-});
-process.on('uncaughtException', err => {
-  exitWithError(err);
-});
-
 const jambo = buildJamboCLI(process.argv);
-jambo && jambo.parse();
+jambo && jambo.parseAsync().catch(err => {
+  // Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
+  exitWithError(err);
+});
--- a/src/commands/build/pagesetsbuilder.js
+++ b/src/commands/build/pagesetsbuilder.js
@@ -6,6 +6,7 @@ const PageSet = require('../../models/pageset');
 const PageTemplate = require('../../models/pagetemplate');
 const PageTemplateDirector = require('./pagetemplatedirector');
 const { NO_LOCALE } = require('../../constants');
+const { warn } = require('../../utils/logger');
 
 /**
  * PageSetsBuilder is responsible for matching {@link PageConfigs} and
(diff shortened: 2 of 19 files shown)


oshi97:
image
image
image
image
image

src/commands/build/pagesetsbuilder.js
```diff
@@ -47,7 +48,7 @@ module.exports = class PageSetsBuilder {
         const localeMessage = locale !== NO_LOCALE
           ? ` for '${locale}' locale`
           : '';
-        console.log(
+        warn(
           `Warning: No page templates found${localeMessage}, not generating a ` +
```
cea2aj, line 52: nit: if the warning message says 'warn', then maybe we don't need 'Warning' at the start of the error message

oshi97: ah yeah that's a good point

oshi97:
image

oshi97:
jambo help still prints normally
image

oshi97:
jambo --version prints normally
image

src/utils/errorutils.js
```diff
@@ -1,18 +1,15 @@
...
  * async operations will be lost.
  * @param {Error} error
  */
-exports.exitWithError = (err) => {
-  fs.writeSync(process.stderr.fd,
-    `${err.message}\n` +
-    `${err.stack}`
-  );
+exports.exitWithError = (err = {}) => {
+  err.stack && error(err.stack);
```
cea2aj, line 12: Do you know if the error writing is synchronous? I used fs.writeSync here previously because I read somewhere that console.log() was async, and that could mean that the process could exit before the error is printed. After taking another look, it looks like node updated console.log() to be synchronous so that's probably no longer an issue

cea2aj, line 12: Also, can we include the error message. I believe there were some messages which don't contain the message in the stack, so if we remove the message we could be losing some information

src/cli.js
```diff
@@ -2,13 +2,8 @@
...
 const buildJamboCLI = require('./buildJamboCLI');
 
-// Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
-process.on('unhandledRejection', err => {
-  exitWithError(err);
-});
-process.on('uncaughtException', err => {
+const jambo = buildJamboCLI(process.argv);
+jambo && jambo.parseAsync().catch(err => {
+  // Exit with a non-zero exit code for unhandled rejections and uncaught exceptions
```
cea2aj, line 7: does this really catch all unhandled rejections and uncaught exceptions?

src/utils/errorutils.js
oshi97: do you have an example? I tried throwing the regular Error, SystemError, and UserError in cli.js and the stack always contained the message. It looks like Error always contains the message in the stack? https://nodejs.org/api/errors.html#errors_error_message
I was seeing the message duplicated when I tried printing both the message and the stack, and I also see this behavior if I use the current develop version of jambo, so I think we should be good to not double print message?
image

src/cli.js
oshi97: It can miss exceptions/rejections that occur during or before parseAsync and buildJamboCLI so it does miss a few! updated

oshi97:
what it looks like on cream terminal
image

src/utils/errorutils.js
cea2aj: When a UserError or SystemError is created with both a message and a stack trace, the stack will only contain the inner message and not the one that was supplied when the error was created. One example is adding an extra comma somewhere in the jambo.json so that it breaks. The current setup allows us to print both messages: 
```
UserError: Error parsing jambo.json
SyntaxError: Unexpected token ,
    at unexpected (/Users/canderson/jambo/node_modules/comment-json/src/parse.js:81:17)
    at expect (/Users/canderson/jambo/node_modules/comment-json/src/parse.js:126:5)
    at parse_object (/Users/canderson/jambo/node_modules/comment-json/src/parse.js:260:5)
```

One idea is that we can modify UserError and SystemError so that any provided message is added to the front of the stack trace

oshi97: that seems to work!

oshi97:
jambo errors pass through the original stack correctly
image

cea2aj (approved):
lgtm

PR #229, opened 2021-06-07
Add Jambo build validation hook

yen-tt:
- Added another operation inside PageWriter to ensure template data is correct
using TemplateDataValidator, which pass in the formatted config data to the theme's validation
hook function

J=SLAP-1369
TEST=auto

Created Jest tests using default config format with/without missing field when
pass to validation hook and check if errors are handle properly

### DIFF
--- a/src/commands/build/pagewriter.js
+++ b/src/commands/build/pagewriter.js
@@ -7,6 +7,7 @@ const UserError = require('../../errors/usererror');
 const { NO_LOCALE } = require('../../constants');
 const LocalizationConfig = require('../../models/localizationconfig');
 const TemplateArgsBuilder = require('./templateargsbuilder');
+const TemplateDataValidator = require('./templatedatavalidator');
 
 /**
  * PageWriter is responsible for writing output files for the given {@link PageSet} to
@@ -29,6 +30,12 @@ module.exports = class PageWriter {
      */
     this._templateArgsBuilder =
       new TemplateArgsBuilder(config.templateDataFormatterHook);
+
+    /**
+     * @type {TemplateDataValidator}
+     */
+     this._templateDataValidator =
+      new TemplateDataValidator(config.templateDataValidationHook);
   }
 
   /**
@@ -50,7 +57,6 @@ module.exports = class PageWriter {
         throw new UserError(`Error: No config found for page: ${page.getName()}`);
       }
 
-      console.log(`Writing output file for the '${page.getName()}' page`);
       const templateArguments = this._templateArgsBuilder.buildArgs({
         relativePath: this._calculateRelativePath(page.getOutputPath()),
         pageName: page.getName(),
@@ -60,7 +66,13 @@ module.exports = class PageWriter {
         locale: pageSet.getLocale(),
         env: this._env
       });
+      
+      this._templateDataValidator.validate({
+        pageName: page.getName(),
+        pageData: templateArguments
+      });
 
+      console.log(`Writing output file for the '${page.getName()}' page`);
       const template = hbs.compile(page.getTemplateContents());
       const outputHTML = template(templateArguments);
 
--- a/src/commands/build/pagewriter.js
+++ b/src/commands/build/pagewriter.js
@@ -43,6 +43,7 @@ module.exports = class PageWriter {
    * Writes a file to the output directory per page in the given PageSet.
    *
    * @param {PageSet} pageSet the collection of pages to generate
+   * @throws {UserError} on missing config(s)
    */
   writePages(pageSet) {
     if (!pageSet || pageSet.getPages().length < 1) {
@@ -68,10 +69,12 @@ module.exports = class PageWriter {
         env: this._env
       });
       
-      this._templateDataValidator.validate({
+      if(!this._templateDataValidator.validate({
         pageName: page.getName(),
         pageData: templateArguments
-      });
+      })) {
+        throw new UserError('Invalid page template configuration(s).');
+      }
 
       info(`Writing output file for the '${page.getName()}' page`);
       const template = hbs.compile(page.getTemplateContents());
--- a/src/commands/build/pagewriter.js
+++ b/src/commands/build/pagewriter.js
@@ -35,7 +35,7 @@ module.exports = class PageWriter {
     /**
      * @type {TemplateDataValidator}
      */
-     this._templateDataValidator =
+    this._templateDataValidator =
       new TemplateDataValidator(config.templateDataValidationHook);
   }
 
@@ -43,7 +43,8 @@ module.exports = class PageWriter {
    * Writes a file to the output directory per page in the given PageSet.
    *
    * @param {PageSet} pageSet the collection of pages to generate
-   * @throws {UserError} on missing config(s)
+   * @throws {UserError} on missing page config(s), validation hook execution 
+   * failure, and invalid template data using Theme's validation hook
    */
   writePages(pageSet) {
     if (!pageSet || pageSet.getPages().length < 1) {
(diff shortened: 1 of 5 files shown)


src/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,41 @@
...
+const { info } = require('../../utils/logger');
+
+/**
+ * TemplateDataValidator is reponsible for checking data supplied to a page
+ * using Theme's custom validation steps (if any).
+ */
+module.exports = class TemplateDataValidator {
+  constructor(templateDataValidationHook) {
+    /**
+     *  The path to template data validation hook.
```
cea2aj, line 13: nit: we don't usually have a space in front of the description

src/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,41 @@
...
+     * @type {string}
+     */
+    this._templateDataValidationHook = templateDataValidationHook;
+  }
+  
+  /**
+   *  Execute validation hook's function if file exists
```
cea2aj, line 20: nit: extra space here as well

src/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,41 @@
...
+     */
+    this._templateDataValidationHook = templateDataValidationHook;
+  }
+  
+  /**
+   * Execute validation hook's function if file exists
+   * @param {string} pageName name of the current page
+   * @param {Object} pageData template arguments for the current page
+   */
+  validate({ pageName, pageData }) {
```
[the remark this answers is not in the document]
cea2aj: We haven't followed it everywhere in this repo, but I think it makes sense to use it going forward

tests/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,83 @@
...
+        global_config,
+        params,
+        relativePath,
+        env
+    };
+
+    expect(() => new TemplateDataValidator(templateDataValidationHook).validate({
+        pageName: 'examplePage',
+        pageData: templateData
+    })).not.toThrow(UserError);
```
cea2aj, line 61: what do think about just checking that an 'Error' isn't thrown? That way the test will catch it if something other than a UserError is thrown

src/commands/build/templatedatavalidator.js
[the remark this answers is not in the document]
yen-tt: It's a little hard to define this param pageData since it could be structured using the default template data format or using Theme's custom formatter.

tests/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,83 @@
...
+        global_config_missing_key,
+        params,
+        relativePath,
+        env
+    };  
+
+    expect(() => new TemplateDataValidator(templateDataValidationHook).validate({
+        pageName: 'examplePage',
+        pageData: templateData
+    })).toThrow(UserError);
```
cea2aj, line 80: I think using 'UserError' makes sense here though because we are testing a specific error

src/commands/build/templatedatavalidator.js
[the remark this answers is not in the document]
cea2aj: oh, we don't need to define the whole object. I was thinking something like this:

```
/**
   * Execute validation hook's function if file exists
   * @param {Object} obj
   * @param {string} obj.pageName name of the current page
   * @param {Object} obj.pageData template arguments for the current page
   */
```

src/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,41 @@
...
+    if (!fs.existsSync(this._templateDataValidationHook)) {
```
tmeyer2115, line 26: Small nit: I think we could optimize this a bit. We really only need to do this check once, on construction. If we called `validate` multiple times, it would result in duplicated `existsSync` calls.

src/commands/build/templatedatavalidator.js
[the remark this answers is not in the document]
tmeyer2115: Could we also indicate that this method can throw a `UserError`?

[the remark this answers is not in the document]
yen-tt: ah ok, will do.

src/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,41 @@
...
+        return;
+    }
+    try {
+        info(`Validating configuration for page "${pageName}".`);
+        const validatorFunction = require(this._templateDataValidationHook);
+        validatorFunction(pageData);
```
tmeyer2115, line 32: I'm wondering if we want this class and the hook to throw an exception when the data is invalid. The alternative is that both return a boolean. I think the boolean has its advantages:
- Consumers of the validators can more easily decide how they want to handle bad templates. It may be that they thrown an exception if the validation result is `false`. Or, they try and handle more gracefully. As an example, in the future, maybe users can configure Jambo to gracefully ignore bad pages and continue generation (although I think the more catastrophic failure is the way to go for now). 
- We can differentiate between an actual issue running the Theme supplied validator and an invalid template. Right now, if there was an exception thrown because the validator couldn't run properly, we would treat that the same as the validator finding the template to be no good. We should probably treat the two situations differently and give different errors.

yen-tt: That makes sense, it would give user more flexibility on exception handling. I'm thinking of having the hook returns a boolean, like you suggested, and use that for the validator class to determines whether it should throw the exception or not. For the second point, I believe you can differentiate the two issues right now. Assuming the theme would provide a custom error (UserError) for any template issues, the validator would throw that accordingly. Otherwise, it must be something wrong with running the hook itself so validator should throw a new error with a fail to execute message.

yen-tt: That makes sense, it would give user more flexibility on exception handling. I'm thinking of having the hook returns a boolean, like you suggested, and have this class use the boolean to determines whether or not it should throw an exception. This would fall on the Theme’s hook to log in detail what is invalid, and the validator will throw a new error of general bad template if hook return true.

src/commands/build/pagewriter.js
```diff
@@ -30,12 +31,19 @@ module.exports = class PageWriter {
      */
     this._templateArgsBuilder =
       new TemplateArgsBuilder(config.templateDataFormatterHook);
+
+    /**
+     * @type {TemplateDataValidator}
+     */
+     this._templateDataValidator =
+      new TemplateDataValidator(config.templateDataValidationHook);
```
tmeyer2115, line 39: Is this line indented properly? Doesn't seem to match the indentation of line 33

src/commands/build/pagewriter.js
```diff
@@ -30,12 +31,19 @@ module.exports = class PageWriter {
...
   }
 
   /**
    * Writes a file to the output directory per page in the given PageSet.
    *
    * @param {PageSet} pageSet the collection of pages to generate
+   * @throws {UserError} on missing config(s)
```
tmeyer2115, line 46: Small nit: It's not so much that the error is thrown for missing config(s) but for template data that fails a Theme's validation. In theory, it could fail for any number of reasons: missing keys, improperly structured data, etc.

src/commands/build/templatedatavalidator.js
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Whether or not the file for template data validation hook exists
+     * @type {boolean}
+     */
+    this._hasHook = fs.existsSync(this._templateDataValidationHook);
+  }
+  
+  /**
+   * Execute validation hook's function if file exists
+   * @param {Object} page
```
tmeyer2115, line 27: Newline between 26 and 27

src/commands/build/pagewriter.js
yen-tt: Added that to address existing code (line 59) where it checks for config file, but forgot to include the new errors throw from the validation. Going to update that

tmeyer2115 (approved):

PR #233, opened 2021-06-23
add --globs param to extract-translations cmd

oshi97:
This commit updates the extract-translations command
to take in a --globs argument.
The TranslationExtractor was refactored to only take in
a glob, instead of taking in arrays of files, directories,
and ignored files. The logic of creating an array of globs
by reading from the jambo.json and gitignore files has
been moved into the DefaultTranslationGlobber.

simple-git was updated from 1.131 to 2.40 to remove git
ENOENT errors during acceptance tests. There were improvements
to how simple-git performs git operations in parallel after v2.
We only use simple-git for very basic operations like cloning,
init-ing, and checking out submodules, staging files, and committing,
so there shouldn't be any issues with upgrading. The only breaking
changes in v2.0.0 are git.log using strict ISO by default, and changes
to methods that being with underscores (aka private methods).

J=SLAP-1378
TEST=manual,auto

run jambo extract-translations in the theme, and also in campbells-ca
see messages.pot generated as expected

add acceptance + unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "merge-options": "^2.0.0",
     "npmlog": "^4.1.2",
     "prompts": "^2.3.1",
-    "simple-git": "^2.40.0",
+    "simple-git": "^1.131.0",
     "yargs": "^17.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "merge-options": "^2.0.0",
     "npmlog": "^4.1.2",
     "prompts": "^2.3.1",
-    "simple-git": "^1.131.0",
+    "simple-git": "^2.40.0",
     "yargs": "^17.0.0"
   },
   "devDependencies": {
--- a/src/commands/extract-translations/defaulttranslationglobber.js
+++ b/src/commands/extract-translations/defaulttranslationglobber.js
@@ -1,0 +1,50 @@
+const fs = require('fs');
+
+module.exports = class DefaultTranslationGlobber {
+  constructor(dirs = {}, ignoredPaths = []) {
+    this.pages = dirs.pages || '';
+    this.partials = dirs.partials || [];
+    this.ignoredPaths = ignoredPaths;
+    this.extensions = ['.js', '.hbs']; // only extract from files with these extensions
+  }
+
+  getGlobs() {
+    const { files, directories } = this._getFilesAndDirsFromJamboConfig();
+    return this._globInputFilePaths(directories, files, this.ignoredPaths);
+  }
+
+  /**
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "merge-options": "^2.0.0",
     "npmlog": "^4.1.2",
     "prompts": "^2.3.1",
-    "simple-git": "^1.131.0",
+    "simple-git": "^2.40.0",
```
tmeyer2115, line 41: Any worries about the major version upgrade?

src/commands/extract-translations/defaulttranslationglobber.js
```diff
@@ -0,0 +1,50 @@
+const fs = require('fs');
+
+module.exports = class DefaultTranslationGlobber {
```
tmeyer2115, line 3: JSDoc

src/commands/extract-translations/defaulttranslationglobber.js
```diff
@@ -0,0 +1,50 @@
...
+   */
+   _globInputFilePaths(directories, specificFiles, ignoredPaths) {
+    const extensions = this.extensions.join(',');
+    const directoryGlobs = directories.map(dirpath => `${dirpath}/**/*{${extensions}}`);
+    const ignoreGlobs = ignoredPaths.map(dirpath => `!${dirpath}`);
+    return [...directoryGlobs, ...specificFiles, ...ignoreGlobs];
+  }
+
+  /**
+   * Returns an array of files and array of directories contained in the jamboConfig.
```
tmeyer2115, line 32: Can we be more specific in the JSDoc here and in the method name? We're not just pulling any old files and dirs, we're accessing the site-specific partials/templates.

package.json
oshi97: ah shoot I added a blurb in the commit msg after rebasing and amending, but didn't update the github commit msg
full blurb below, but basically the only breaking change was git.log using strict ISO by default, and changes to
methods that begin with underscores. We only use very basic functionality so we should be good.

_simple-git was updated from 1.131 to 2.40 to remove git
ENOENT errors during acceptance tests. There were improvements
to how simple-git performs git operations in parallel after v2.
We only use simple-git for very basic operations like cloning,
init-ing, and checking out submodules, staging files, and committing,
so there shouldn't be any issues with upgrading. The only breaking
changes in v2.0.0 are git.log using strict ISO by default, and changes
to methods that being with underscores (aka private methods)._

oshi97: oddly enough it seems like the tests DO still pass though
image

tmeyer2115 (approved):

PR #235, opened 2021-06-28
Changed upgrade functionality

cjiang2000:
Changed upgrade functionality, so that now it reverts to old theme files when it errors out.

J=SLAP-1308
TEST=manual/automatic

Tested by running jambo upgrade on the hitchhiker theme and making sure that upgrade ran unchanged when no error occurs. Also ran with an artificial error to make sure that theme files are reverted correctly.

Created acceptance test for themeupgrader that checks to make sure that when jambo upgrade errors out, the test-theme remains unchanged.

### DIFF
--- a/src/commands/upgrade/themeupgrader.js
+++ b/src/commands/upgrade/themeupgrader.js
@@ -77,8 +77,8 @@ class ThemeUpgrader {
     }
   }
 
-  execute(args) {
-    this._upgrade({
+  async execute(args) {
+    await this._upgrade({
       themeName: this.jamboConfig.defaultTheme,
       disableScript: args.disableScript,
       isLegacy: args.isLegacy,
@@ -104,12 +104,20 @@ class ThemeUpgrader {
       throw new UserError(
         `Theme "${themeName}" not found within the "${this._themesDir}" folder`);
     }
-    
-    if (await this._isGitSubmodule(themePath))  {
+    if (await this._isGitSubmodule(themePath)) {
       await this._upgradeSubmodule(themePath, branch)
     } else {
-      await this._recloneTheme(themeName, themePath, branch);
-      this._removeGitFolder(themePath);
+      const tempDir = fs.mkdtempSync('./');
+      try { 
+        fs.copySync(themePath, tempDir);
+        await this._recloneTheme(themeName, themePath, branch);
+        this._removeGitFolder(themePath);
+        fs.removeSync(tempDir);
+      }
+      catch (error) {
+        fs.moveSync(tempDir, themePath);
+        throw error;
+      }
     }
     if (!disableScript) {
       this._executePostUpgradeScript(themePath, isLegacy);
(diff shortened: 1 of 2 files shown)


src/commands/upgrade/themeupgrader.js
```diff
@@ -125,71 +133,72 @@ class ThemeUpgrader {
...
+ * Returns whether the given file path is registered as a git submodule.
+ * @param {string} submodulePath
+ * @returns {boolean}
+ */
+async _isGitSubmodule(submodulePath) {
+  const submodulePaths = await git.subModule(['foreach', '--quiet', 'echo $sm_path']);
+  return !!submodulePaths
+    .split('\n')
+    .find(p => p === submodulePath)
+}
```
cea2aj, line 201: I think you accidentally removed some whitespace in front of these lines

src/commands/upgrade/themeupgrader.js
```diff
@@ -104,12 +104,20 @@ class ThemeUpgrader {
...
-    
-    if (await this._isGitSubmodule(themePath))  {
+    if (await this._isGitSubmodule(themePath)) {
       await this._upgradeSubmodule(themePath, branch)
     } else {
-      await this._recloneTheme(themeName, themePath, branch);
-      this._removeGitFolder(themePath);
+      const tempDir = fs.mkdtempSync('./');
+      try { 
+        fs.copySync(themePath, tempDir);
```
cea2aj, line 112: I suppose the `fs.copySync` could fail, in which case the revert would not work properly, however I'm not sure if there's a simple way to address that

cea2aj: I think there's a pretty low risk of that failing since we check that the theme path exists, so you can ignore these comments

tests/acceptance/suites/themeupgrader.js
```diff
@@ -0,0 +1,27 @@
+const process = require('process');
+const git = require('simple-git/promise')();
+const { runInPlayground } = require('../setup/playground');
+const ThemeManager = require('../../../src/utils/thememanager');
+const path = require('path');
+
+ThemeManager.getRepoForTheme = () => {
```
cea2aj, line 7: Do we need to mock this if we are passing in the theme URL in the jambo import command?

tests/acceptance/suites/themeupgrader.js
```diff
@@ -0,0 +1,27 @@
...
+    return path.resolve(__dirname, '../test-themes/basic-flow');
+};
+
+const error = console.error;
+console.error = jest.fn();
+afterAll(() => console.error = error);
```
cea2aj, line 13: Can you add a comment to explain why this was needed? Is it because the upgrade branch fail would otherwise output a bunch of error messages when running tests?

tests/acceptance/suites/themeupgrader.js
oshi97: Yup! the mock is for the upgrade command which also uses getRepoForTheme

src/commands/upgrade/themeupgrader.js
cjiang2000: I'm not sure where exactly you're referring to. Could you be more specific?

tests/acceptance/suites/themeupgrader.js
cjiang2000: __recloneTheme also calls ThemeManager.getRepoForTheme(), so we still need the mock

tests/acceptance/suites/themeupgrader.js
cjiang2000: upgrade --branch fail outputs an error message when we run tests. The tests still pass, we just thought it would be cleaner if they didn't display error messages when we run them.
I'll specify that it's to silence errors.

src/commands/upgrade/themeupgrader.js
cjiang2000: Jk, it's easier to see in vscode, i just fixed the tabs

cea2aj (approved):
LGTM

PR #236, opened 2021-08-02
locale fallback acceptance test

yen-tt:
Created locale fallback acceptance test

- created a new test-theme folder `locale-fallback` with locale_config.json that contains 3 locales: en (default locale), fr (locale fallback for es), and es.
- 'locale-fallback' acceptance test: use `locale-fallback` theme and create an index page for en and es, `index.html.hbs` and `index.fr.html.hbs,` using two different templates. When executing jambo page command for en index page, also create a pageConfig for es with the flag `--locales es`. See that es index page fall back on the pageTemplate and config file for fr.

J=SLAP-1303
TEST=auto

Set `procrastinateCleanup` to true when running the test. See that in the playground folder, the expected config files, pageTemplates, and public html files are build as expected for en, es, and fr. See that acceptance test passed

### DIFF
--- a/tests/acceptance/test-themes/locale-fallback/postimport.js
+++ b/tests/acceptance/test-themes/locale-fallback/postimport.js
@@ -1,0 +1,23 @@
+#!/usr/bin/env node
+const fs = require('fs');
+const fsExtra = require('fs-extra');
+const { assign, stringify } = require('comment-json');
+
+/**
+ * Updates the defaultTheme field in the jambo.json.
+ * 
+ * @param {string} themeName 
+ */
+function updateDefaultTheme(themeName) {
+  const jamboConfig = JSON.parse(fs.readFileSync('jambo.json', 'utf-8'));
+  if (jamboConfig.defaultTheme !== themeName) {
+    const updatedConfig = assign({ defaultTheme: themeName }, jamboConfig);
+    fs.writeFileSync('jambo.json', stringify(updatedConfig, null, 2));
+  }
+}
+
+updateDefaultTheme('locale-fallback');
+fs.writeFileSync('config/global_config.json', '{}');
+fsExtra.copySync('themes/locale-fallback/config/locale_config.json', 
+  'config/locale_config.json');
+fsExtra.copySync('themes/locale-fallback/static', 'static');
--- a/tests/acceptance/test-themes/locale-fallback/postimport.js
+++ b/tests/acceptance/test-themes/locale-fallback/postimport.js
@@ -11,7 +11,7 @@ const { assign, stringify } = require('comment-json');
 function updateDefaultTheme(themeName) {
   const jamboConfig = JSON.parse(fs.readFileSync('jambo.json', 'utf-8'));
   if (jamboConfig.defaultTheme !== themeName) {
-    const updatedConfig = assign({ defaultTheme: themeName }, jamboConfig);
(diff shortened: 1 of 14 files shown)


tests/acceptance/test-themes/locale-fallback/postimport.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Updates the defaultTheme field in the jambo.json.
+ * 
+ * @param {string} themeName 
+ */
+function updateDefaultTheme(themeName) {
+  const jamboConfig = JSON.parse(fs.readFileSync('jambo.json', 'utf-8'));
+  if (jamboConfig.defaultTheme !== themeName) {
+    const updatedConfig = assign({ defaultTheme: themeName }, jamboConfig);
```
cea2aj, line 14: I would expect this assign call to look like: `assign({}, jamboConfig, { defaultTheme: themeName });` because assign operates right to left. If left as-is, wouldn't the defaultTheme from the jamboConfig take precedence over the themeName specified in the function parameter?

tests/acceptance/test-themes/locale-fallback/postimport.js
```diff
@@ -0,0 +1,23 @@
...
+    fs.writeFileSync('jambo.json', stringify(updatedConfig, null, 2));
+  }
+}
+
+updateDefaultTheme('locale-fallback');
+fs.writeFileSync('config/global_config.json', '{}');
+fsExtra.copySync('themes/locale-fallback/config/locale_config.json', 
+  'config/locale_config.json');
+fsExtra.copySync('themes/locale-fallback/static', 'static');
```
cea2aj, line 23: I'm curious, what do we need to copy the static folder for?

cea2aj:
Maybe I'm missing something, but why does the fallback happen in the first place? I thought that we fallback if config options or pages are missing, but aren't those configs and pages present?

tests/acceptance/test-themes/locale-fallback/postimport.js
yen-tt: sitegenerator.js build function expects static folder from theme and the test site folder when creating static output directory (this._createStaticOutput(staticDirs, config.dirs.output)) or it will throw an error

tests/acceptance/test-themes/locale-fallback/postimport.js
yen-tt: In this test, since jamboConfig is the default jambo.json config file generated from the init command, there's no defaultTheme so it wouldn't override it. This function is follow the same format as the other acceptance tests. But I think your assign call make sense if jamboConfig is somehow alter before this postimport call in new acceptance tests. I could update this and other test's postimport if you think that's better?

yen-tt: Updated! just a note, the [assign method](https://openbase.com/js/comment-json/documentation#assigntarget-object-source-object-keys-array) used here is from comment-json which have different params from Object.assign. Instead of having multiple sources like Object.assign, the third param in this assign function takes an array of specific keys to copy.

cea2aj: ya, writing into jamboConfig should be fine in this case

cea2aj (approved):

PR #238, opened 2021-08-17
boilerplate for typescript migration

oshi97:
The CI is expected to not pass here. Source code changes necessary to get the build + tests working will come in an following PR.

1. use cjs-to-es6 to change all imports/exports from require() to import, a number of requires() and module.exports/exports had to be manually patched due to us using less typical syntax
2. renames all files under src from .js to .ts, and adds tsconfig. Since we restrict to node 12 and up, we can use target: "es2019", lib: "es2019" for things like array.prototype.flatMap (which we already use)
3. adds babel.config.js so that jest can understand es6 and typescript, and updates eslint config
4. adds npm dependencies like `@handlebars/parser`, babel-jest, and other typescriptey things (also updates yargs + jest),

J=SLAP-1491
TEST=manual

smoke test jambo (with the upcoming PRs included)


.eslintrc.json
```diff
@@ -1,25 +1,51 @@
...
+    ],
+    "quote-props": [
+      "error",
+      "as-needed"
+    ],
+    "max-len": [
+      "error",
+      {
+        "code": 125,
+        "ignorePattern": "^const\\s.+=\\s*require\\s*\\("
```
cea2aj, line 39: I think this should be updated since we are now using import

package.json
```diff
@@ -2,21 +2,26 @@
...
     "url": "git+https://github.com/yext/jambo.git"
   },
   "bin": {
-    "jambo": "src/cli.js"
+    "jambo": "lib/cli.js"
   },
   "engines": {
     "node": ">=12.0.0"
   },
-  "author": "nbramblett@yext.com",
```
cea2aj, line 19: lol

oshi97: gotem

src/handlebars/registerhbshelpers.ts
```diff
@@ -34,11 +34,11 @@ module.exports = function registerHbsHelpers(hbs) {
     gte: function(v1, v2) {
       return v1 >= v2;
     },
-    and: function() {
-      return Array.prototype.slice.call(arguments).every(Boolean);
+    and: function(...args) {
+      return Array.prototype.slice.call(args).every(Boolean);
```
cea2aj, line 38: for my own edification, how was this working before?

.eslintrc.json
oshi97: I just ended up removing the ignore pattern, I can add it back in a later PR if need be

src/handlebars/registerhbshelpers.ts
oshi97: javascript has a special `arguments` keyword 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments
typescript-eslint/recommended eslint was complaining about using it, though

cea2aj:
It's a little odd to me that #239 is showing up in this diff. Would that go away if we merged #239 into feature/typescript-migration first? Otherwise, LGTM

oshi97:
> It's a little odd to me that #239 is showing up in this diff. Would that go away if we merged #239 into feature/typescript-migration first? Otherwise, LGTM

239 has already been merged in (by merging develop into this branch), I think I can fix it by rebasing though

cea2aj:
> > It's a little odd to me that #239 is showing up in this diff. Would that go away if we merged #239 into feature/typescript-migration first? Otherwise, LGTM
> 
> ~239 has already been merged in (by merging develop into this branch), I think I can fix it by rebasing though~
> ah nevermind I thought you meant merge into the dev branch not feature branch, I think that would have worked
> I rebased both the dev and feature branch but it still shows up as changed, I think this is a bug in github because if I make a new pull request with the same 2 branches in the changed files it displays as expected

Alright, I think we're fine then. When we merge this feature branch into develop I think that part of the diff will go away

cea2aj (approved):

PR #239, opened 2021-08-18
update canonicalizeLocale to handle 中文(chinese)

oshi97:
J=SLAP-1526
TEST=auto,manual

unit tests
ran build with zh-hans-ch in locale config, saw that the generated source code used the correct locale in the page
tried building a page with a locale with >3 sections, got a jambo warning but build continued without changing the locale

### DIFF
--- a/src/utils/i18nutils.js
+++ b/src/utils/i18nutils.js
@@ -12,19 +12,16 @@ canonicalizeLocale = function(localeCode) {
   }
   const localeCodeSections = localeCode.replace(/-/g, '_').split('_');
   const numSections = localeCodeSections.length;
+  const language = localeCodeSections[0].toLowerCase();
   if (numSections === 1) {
-    return localeCodeSections[0].toLowerCase();
+    return language;
   } else if (numSections === 2) {
-    const language = localeCodeSections[0].toLowerCase();
-    const region = localeCodeSections[1].toUpperCase();
-    return `${language}_${region}`; 
+    if (language === 'zh') {
+      return formatLocale(language, localeCodeSections[1]);
+    }
+    return formatLocale(language, null, localeCodeSections[1]);
   } else if (numSections === 3) {
-    const language = localeCodeSections[0].toLowerCase();
-    const rawModifier = localeCodeSections[1];
-    const languageModifier =
-      rawModifier.charAt(0).toUpperCase() + rawModifier.slice(1).toLowerCase();
-    const region = localeCodeSections[2].toUpperCase();
-    return `${language}-${languageModifier}_${region}`;
+    return formatLocale(language, localeCodeSections[1], localeCodeSections[2]);
   } else if (numSections > 3) {
     warn(
       `Encountered strangely formatted locale "${localeCode}", ` +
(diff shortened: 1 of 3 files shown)


tests/acceptance/suites/themeupgrader.js
```diff
@@ -5,7 +5,7 @@ const ThemeManager = require('../../../src/utils/thememanager');
 const path = require('path');
 
 ThemeManager.getRepoForTheme = () => {
-    return path.resolve(__dirname, '../test-themes/basic-flow');
+  return path.resolve(__dirname, '../test-themes/basic-flow');
```
oshi97, line 8: this test failed occasionally locally due to it taking >5s as an async test, otherwise I would not have noticed the indentation being off

oshi97:
was testing in the theme and this doesn't handle `zh-Hans` correctly (i.e. chinese w/ no region code)

src/utils/i18nutils.js
```diff
@@ -8,18 +10,26 @@ canonicalizeLocale = function(localeCode) {
...
-
-  if (localeCodeSections.length > regionIndex) {
-    localeCodeSections[regionIndex] = localeCodeSections[regionIndex].toUpperCase();
+  const localeCodeSections = localeCode.replace(/-/g, '_').split('_');
+  const numSections = localeCodeSections.length;
+  if (numSections === 1) {
+    return localeCodeSections[0].toLowerCase();
+  } else if (numSections === 2) {
+    const language = localeCodeSections[0].toLowerCase();
+    const region = localeCodeSections[1].toUpperCase();
```
cea2aj, line 19: Someone could pass in `zh-Hant` which would be two sections, but the second one would be the script and not the language. Maybe we should have list of known scripts? If it the second section matches one of the known language scripts then we know its a region instead. Alternatively we say that if the second section has 4 chars, its a language script, otherwise if it has two characters its a region

oshi97: yup I realized this too, I just added a case for whether the language is chinese.

cea2aj (approved):

src/utils/i18nutils.js
```diff
@@ -1,25 +1,70 @@
...
+/**
+ * Parses a locale code into its constituent parts.
+ * 
+ * @param {string} localeCode 
+ * @returns { language: string, modifier?: string, region?: string } 
+ */
+function parseLocale(localeCode) {
+  const localeCodeSections = localeCode.replace(/-/g, '_').split('_');
+  const numSections = localeCodeSections.length;
+  const language = localeCodeSections[0].toLowerCase();
```
cea2aj, line 26: nit: should we also capitalize region here and make the modifer uppercase?

oshi97: yeah I can write some tests for the intermediary methods too

src/utils/i18nutils.js
```diff
@@ -16,39 +16,55 @@ exports.canonicalizeLocale = function(localeCode) {
...
   const language = localeCodeSections[0].toLowerCase();
-  if (numSections === 1) {
-    return { language };
-  } else if (numSections === 2) {
-    if (language === 'zh') {
+  const parseModifierAndRegion = () => {
+    const numSections = localeCodeSections.length;
+    if (numSections === 1) {
+      return {};
+    } else if (numSections === 2 && language === 'zh') {
```
cea2aj, line 31: Should we handle codes such as: `zh_CH`? Currently it would treat `CH` as a modifier when really its a region

oshi97: yeah I missed this case

cea2aj (approved):
🔥🔥🔥

PR #240, opened 2021-08-19
use es6 dynamic import for hbs helpers

oshi97:
jambo build is already an async command, so its easy to refactor
this dynamic require into a dynamic import statement.
The other dynamic requires in jambo are harder to refactor because
they live inside fully synchronous commands/methods, like our custom
commands importer.

TEST=manual

smoke test jambo (on a different branch that has all TS changes)

### DIFF
--- a/tests/fixtures/hooks/templatedataformatter.js
+++ b/tests/fixtures/hooks/templatedataformatter.js
@@ -11,7 +11,7 @@
  * 
  * @param {Object} pageNameToConfig object of pageName to pageConfig
  */
-export default function (pageMetadata, siteLevelAttributes, pageNameToConfig) {
+module.exports = function (pageMetadata, siteLevelAttributes, pageNameToConfig) {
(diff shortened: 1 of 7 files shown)


tests/fixtures/hooks/templatedataformatter.js
```diff
@@ -11,7 +11,7 @@
  * 
  * @param {Object} pageNameToConfig object of pageName to pageConfig
  */
-export default function (pageMetadata, siteLevelAttributes, pageNameToConfig) {
+module.exports = function (pageMetadata, siteLevelAttributes, pageNameToConfig) {
```
oshi97, line 14: the fixtures should use commonjs exports because that's currently what jambo expects
it seems like we can also support es6 by looking for both `require(module) || require(module).default` but we don't need to just yet

cea2aj (approved):

PR #241, opened 2021-08-19
fix typo templateDataFormattter -> templateDataFormatterHook

oshi97:
Fixes a typo found by typescript

I was using sublime merge's "stage hunk" feature and the parameter types
were registered in the same hunk as the typo fix, and it didn't seem worth it to manually separate out
the commits.


yen-tt (approved):
Should probably update the jsdoc for all these functions throughout the repos since we have the types now but maybe here or in another pr. but lgtm!

oshi97:
> Should probably update the jsdoc for all these functions throughout the repos since we have the types now, maybe here or in another pr. lgtm!

Yup I think we can do a jsdoc cleaning at the end!

PR #246, opened 2021-08-23
Update to new Command interface

yen-tt:
Update built in commands to use new Command interface

- make `ArgumentMetadata` an interface
- change `ArgumentType` to string union type
- change `Command` class to an interface that enforces static and non static fields/methods
  - Typescript doesn't allow static fields/methods to be used in interface. In order to enforce the static properties in interface (show errors to user if a customcommand is not implemented properly), a workaround is implemented. `CommandExecutable` contains all the non static properties and functions. `Command` holds a constructor signature of type `CommandExecutable` and specified fields (e.g alias, args, and shortdescriptions) that will be required to be static when type checked. This approach is similar to what is used in React (Component and ComponentClass)
  - Command will also take in two type arguments to define the return type of jambo command's arg() method and the argument types that get pass in execute()
  - The syntax to follow is: `const SomeCustomCommand : Command<Args, ExecArgs> = class { ... }` where Args is the type of the args object and ExecArgs is the type of the args passed to execute
 - Keeping the static methods instead of switching to static fields to avoid breaking current public interface and needing to handle another legacy import
 - Update other related files such as commandregistry, commandimporter, and yargsfactory to follow the new interface
 - Update RepositorySettings and PageConfiguration from class to interface
 - remove layout option to pagecommand
 
J=SLAP-1492

TEST=compile & auto

see that tests passed

### DIFF
--- a/src/commands/build/buildcommand.ts
+++ b/src/commands/build/buildcommand.ts
@@ -2,35 +2,26 @@ import { ArgumentMetadata, ArgumentType } from '../../models/commands/argumentme
 import UserError from '../../errors/usererror';
 import { isCustomError } from '../../utils/errorutils';
 import SitesGenerator from './sitesgenerator';
+import Command from '../../models/commands/command';
 
 /**
  * BuildCommand builds all pages in the Jambo repo and places them in the
  * public directory.
  */
-class BuildCommand {
-  sitesGenerator: SitesGenerator
+const BuildCommand : Command = class {
+  static alias = 'build';
+  static shortDescription = 'build the static pages for the site';
+  static args: Record<string, ArgumentMetadata> = {
+    jsonEnvVars: {
+      type: ArgumentType.ARRAY,
+      description: 'environment variables containing JSON',
+      isRequired: false
+    }
+  };
+  sitesGenerator: SitesGenerator;
 
   constructor(sitesGenerator) {
     this.sitesGenerator = sitesGenerator;
-  }
-
-  static getAlias() {
-    return 'build';
-  }
-
-  static getShortDescription() {
-    return 'build the static pages for the site';
-  }
-
-  static args() {
-    return {
-      jsonEnvVars: new ArgumentMetadata({
-        type: ArgumentType.ARRAY,
-        itemType: ArgumentType.STRING, 
-        description: 'environment variables containing JSON',
-        isRequired: false
-      })
-    }
   }
 
   static describe() {
@@ -39,7 +30,7 @@ class BuildCommand {
     }
   }
 
-  async execute(args) {
+  async execute(args: Record<string, any>): Promise<any> {
     try {
       await this.sitesGenerator.generate(args.jsonEnvVars);
     } catch (err) {
--- a/src/commands/build/buildcommand.ts
+++ b/src/commands/build/buildcommand.ts
@@ -14,7 +14,7 @@ const BuildCommand : Command = class {
   constructor(sitesGenerator) {
     this.sitesGenerator = sitesGenerator;
   }
- 
+
   static getAlias() {
     return 'build';
   }
--- a/src/commands/build/buildcommand.ts
+++ b/src/commands/build/buildcommand.ts
@@ -1,14 +1,23 @@
-import { ArgumentMetadataImpl } from '../../models/commands/argumentmetadata';
 import UserError from '../../errors/usererror';
 import { isCustomError } from '../../utils/errorutils';
 import SitesGenerator from './sitesgenerator';
-import Command from '../../models/commands/command';
+import Command, { ArgsForExecute } from '../../models/commands/command';
 
+const args = {
+  jsonEnvVars: {
+    type: 'array',
+    itemType:'string', 
+    description: 'environment variables containing JSON',
+    isRequired: false
+  }
+} as const;
+type Args = typeof args;
+  
 /**
  * BuildCommand builds all pages in the Jambo repo and places them in the
  * public directory.
  */
-const BuildCommand : Command = class {
+const BuildCommand : Command<Args> = class {
   sitesGenerator: SitesGenerator;
 
   constructor(sitesGenerator) {
@@ -24,14 +33,7 @@ const BuildCommand : Command = class {
   }
 
   static args() {
-    return {
-      jsonEnvVars: new ArgumentMetadataImpl({
-        type: 'array',
-        itemType:'string', 
-        description: 'environment variables containing JSON',
-        isRequired: false
-      })
-    }
+    return args;
   }
 
   static describe() {
@@ -40,7 +42,7 @@ const BuildCommand : Command = class {
     }
   }
 
-  async execute(args: Record<string, any>): Promise<any> {
+  async execute(args: ArgsForExecute<Args>): Promise<any> {
     try {
       await this.sitesGenerator.generate(args.jsonEnvVars);
     } catch (err) {
--- a/src/commands/build/buildcommand.ts
+++ b/src/commands/build/buildcommand.ts
@@ -12,12 +12,13 @@ const args = {
   }
 } as const;
 type Args = typeof args;
+type ExecArgs = ArgsForExecute<Args>;
   
 /**
  * BuildCommand builds all pages in the Jambo repo and places them in the
  * public directory.
  */
-const BuildCommand : Command<Args> = class {
+const BuildCommand : Command<Args, ExecArgs> = class {
   sitesGenerator: SitesGenerator;
 
   constructor(sitesGenerator) {
@@ -42,7 +43,7 @@ const BuildCommand : Command<Args> = class {
     }
   }
 
-  async execute(args: ArgsForExecute<Args>): Promise<any> {
+  async execute(args: ExecArgs): Promise<any> {
     try {
       await this.sitesGenerator.generate(args.jsonEnvVars);
     } catch (err) {
--- a/src/commands/build/buildcommand.ts
+++ b/src/commands/build/buildcommand.ts
@@ -1,4 +1,4 @@
-import { ArgumentMetadata, ArgumentType } from '../../models/commands/argumentmetadata';
+import { ArgumentMetadata } from '../../models/commands/argumentmetadata';
 import UserError from '../../errors/usererror';
 import { isCustomError } from '../../utils/errorutils';
 import SitesGenerator from './sitesgenerator';
@@ -9,19 +9,29 @@ import Command from '../../models/commands/command';
  * public directory.
  */
 const BuildCommand : Command = class {
-  static alias = 'build';
-  static shortDescription = 'build the static pages for the site';
-  static args: Record<string, ArgumentMetadata> = {
-    jsonEnvVars: {
-      type: ArgumentType.ARRAY,
-      description: 'environment variables containing JSON',
-      isRequired: false
-    }
-  };
   sitesGenerator: SitesGenerator;
 
   constructor(sitesGenerator) {
     this.sitesGenerator = sitesGenerator;
+  }
+ 
+  static getAlias() {
+    return 'build';
+  }
+
+  static getShortDescription() {
+    return 'build the static pages for the site';
+  }
+
+  static args() {
+    return {
+      jsonEnvVars: new ArgumentMetadata({
+        type: 'array',
+        itemType:'string', 
+        description: 'environment variables containing JSON',
+        isRequired: false
+      })
+    }
   }
 
   static describe() {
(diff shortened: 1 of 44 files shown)


cea2aj:
It looks like some of the tests are failing

src/commands/build/buildcommand.ts
```diff
@@ -2,44 +2,35 @@ import { ArgumentMetadata, ArgumentType } from '../../models/commands/argumentme
...
 import SitesGenerator from './sitesgenerator';
+import Command from '../../models/commands/command';
 
 /**
  * BuildCommand builds all pages in the Jambo repo and places them in the
  * public directory.
  */
-class BuildCommand {
-  sitesGenerator: SitesGenerator
+const BuildCommand : Command = class {
```
oshi97, line 11: Is this just another way to write `implements`?

src/commands/import/themeimporter.ts
```diff
@@ -14,47 +14,39 @@ import { searchDirectoryIgnoringExtensions } from '../../utils/fileutils';
...
+const ThemeImporter : Command = class {
   config: JamboConfig;
   _themeShadower: ThemeShadower;
-  _postImportHook: 'postimport'
+  _postImportHook: 'postimport';
+  static alias = 'import';
+  static shortDescription = 'import a theme';
+  static args: Record<string, ArgumentMetadata> = {
+    themeUrl: {
+      type: ArgumentType.STRING,
```
oshi97, line 30: Could we use a string union type for ArgumentType instead of an enum? I think union types are slightly more developer friendly than enums, and have nicer autocomplete/intellisense (also no need to import).

src/commands/import/themeimporter.ts
```diff
@@ -77,7 +69,7 @@ export default class ThemeImporter {
     }
   }
 
-  async execute(args) {
+  async execute(args: Record<string, any>) {
```
oshi97, line 72: we can be more specific here with `Record<string, ArgumentMetadata>`

src/commands/build/buildcommand.ts
```diff
@@ -2,44 +2,35 @@ import { ArgumentMetadata, ArgumentType } from '../../models/commands/argumentme
...
+  static alias = 'build';
```
oshi97, line 12: do static members work on TypeScript interfaces? Like, if I fail to define a `static alias = ` in a command, or set it to the wrong type, will I get an error? I thought this wouldn't work based on https://stackoverflow.com/questions/13955157/how-to-define-static-property-in-typescript-interface but I haven't personally tried anything

src/models/commands/command.ts
```diff
@@ -3,47 +3,48 @@ import { JamboConfig } from '../JamboConfig';
...
-  static getAlias(): string { return ''; }
+  execute(args: Record<string, any>): any
+}
+
+/**
+ * An interface that represents a command in the Jambo CLI. 
+ * Contains non static (CommandExecutable interface) and 
+ * static (specified in here) fields and methods.
+ */
+interface Command {
```
oshi97, line 23: when possible export default should be on the same line as the declaration for the thing being exported
so `export default interface Command` here

src/models/commands/command.ts
```diff
@@ -3,47 +3,48 @@ import { JamboConfig } from '../JamboConfig';
...
+  new(...args: any[]):CommandExecutable;
```
oshi97, line 24: Could you explain this syntax? It's new for me

src/commands/build/buildcommand.ts
yen-tt: sort of, using `implements` interface doesn't allow for static fields/method (From official doc: when a class implements an interface, only the instance side of the class is checked.). This is type checking BuildCommand to conform to Command interface, which contains a constructor signature for an instance and fields/methods belong to the class marked as static

src/commands/import/themeimporter.ts
yen-tt: args here doesn't follow ArgumentMetadata interface (i.e. ArgumentMetadata doesn't have themeUrl, theme, or subModules) so doing so would create error

src/commands/build/buildcommand.ts
yen-tt: Yeah I did some investigation because I ran into that issue with TS interface and static members. This is a workaround I found, which will provide an error if you don't define members like alias as static

src/models/commands/command.ts
yen-tt: sure!

src/models/commands/command.ts
yen-tt: this is defining a constructor signature with any amount of parameters

src/commands/build/buildcommand.ts
oshi97: could we just use public members instead of static?

oshi97: Oh does making them static get around having to instantiate them?

src/commands/import/themeimporter.ts
yen-tt: like so? `type ArgumentType = 'string' | 'number' | 'boolean' | 'array';`
Without using/importing the enum, then user would have to supply the string by themselves rather than selecting whatever is acceptable from ArgumentType enum. 
```
themeUrl: {
      type: 'string'
      ...
```
Error for putting in the wrong type is `Type '"something"' is not assignable to type 'ArgumentType'`.

src/commands/build/buildcommand.ts
yen-tt: Yup, there are areas in the code where we use alias, shortdescription and args before instantiating the command classes or have data to pass to them.

oshi97: I think another option would be exporting a JS object instead of a class, so something like
```ts
export default {
  alias: 'build',
  execute: (jamboConfig, args) => {}
}
``` 
or even something like https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module

oshi97: I'm also okay with the static workaround! It's just a little more complicated that I'd like

yen-tt: for the JS object approach, I guess everything in the class constructor would now just directly pass to the execute function or include some field in Command interface for that? Other helper functions used in execute() is defined in the file but would be not included in the export object, which we type check for Command interface. I'm ok with that as well

cea2aj: I think the solution Yen has here makes sense. If we went with the command module like yargs I think that would be hard to type check. I like the simplicity of the simple JS object rather than a class, however that does raise the question as Yen pointed out of how to handle dependencies of the execute function. We could pass the dependencies to the execute function which seems a little odd. Or we could have a function which acts like a constructor which will create the command objects and include put the dependencies of the execute function on the object so that the execute function can access them. Because of that complexity, I'm leaning towards going with the current solution

src/commands/build/buildcommand.ts
cea2aj: Its just the standard Typescript syntax for specifying a type

src/models/commands/command.ts
```diff
@@ -3,47 +3,48 @@ import { JamboConfig } from '../JamboConfig';
...
   /**
-   * @returns {string} A short, one sentence description of the command. This
-   *                   description appears as part of the help text in the CLI. 
+   * The alias for the command.
    */
-  static getShortDescription(): string {
-    return '';
-  }
-
+  alias: string;
```
cea2aj, line 29: Would we be able to avoid breaking the public interface if we keep these as getter functions?

src/commands/extract-translations/jambotranslationextractor.ts
```diff
@@ -4,52 +4,49 @@ import { info } from '../../utils/logger';
...
-      })
-    };
-  }
-
   static describe() {
-    const args = this.args();
     return {
       displayName: 'Extract Translations',
-      params: Object.keys(args).reduce((params, alias) => {
-        params[alias] = args[alias].toDescribeFormat();
```
cea2aj, line 52: What's the reason for switching away from `toDescribeFormat()`?

cea2aj: Oh, I see. It's because it's now an interface

yen-tt: I left it as that for now since we have an item that will remove these describe functions and have DescribeCommand use args field to sort out the description!

src/models/commands/command.ts
yen-tt: This would help keep must of the current public interface, but we also turn ArgumentMetadata from class to interface so it would be a breaking change either way I think.

src/commands/extract-translations/jambotranslationextractor.ts
oshi97: toDescribeFormat() was also only being used in this command and none of the other commands

src/commands/build/buildcommand.ts
oshi97: (for posterity/record keeping) we ended up deciding on keeping using the static workaround, which still provides nice type checking/intellisense, and also keeping the method signatures the same as the current interfaces for everything
i.e. no interfaces will change

src/models/commands/command.ts
```diff
@@ -1,49 +1,67 @@
...
 
   /**
-   * @returns {Object<string, ArgumentMetadata>} Descriptions of each argument,
-   *                                             keyed by name.
+   * Descriptions of each argument, keyed by name.
    */
-  static args(): Record<string, ArgumentMetadata>{
-    return {};
-  }
+   args(): Record<string, ArgumentMetadata>;
```
tmeyer2115, line 59: What is the relationship between this type and `T`?

tmeyer2115: If `T` is getting transformed according to `ArgsForExecute<T & Record<string, Metadata>>` in `CommandExecutable`, does that mean that `args(): T`?

src/models/commands/argumentmetadatalegacy.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * defines the different kinds of argument that are supported.
+ */
+ type ArgumentType = 'string' | 'number' | 'boolean' | 'array';
```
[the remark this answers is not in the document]
cea2aj: I wanted to keep this separate so that updates to the newer ArguentMetadata model won't change the definition for the legacy model

src/models/commands/argumentmetadatalegacy.ts
```diff
@@ -0,0 +1,41 @@
...
+
+ /**
+  * An interface describing the legacy version of {@link ArgumentMetadata}.
+  */
+ export interface ArgumentMetadataLegacy {
```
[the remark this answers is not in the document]
cea2aj: Yep, there's a `@deprecated` tag

src/models/commands/argumentmetadata.ts
```diff
@@ -1,89 +1,40 @@
...
+ * defines the different kinds of argument that are supported.
  */
-const ArgumentType = {
-  STRING: 'string',
-  NUMBER: 'number',
-  BOOLEAN: 'boolean',
-  ARRAY: 'array'
-}
-Object.freeze(ArgumentType);
+export type ArgumentType = 'string' | 'number' | 'boolean' | 'array';
```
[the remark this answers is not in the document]
cea2aj: I don't believe that's possible because we need to be able to represent those type at runtime. If we switched those from strings to the real types, we'd end up with an object which refers to a type as a value, which doesn't work. Example: `const args = {
  themeUrl: {
    type: string,
    description: 'the url to the theme',
    isRequired: true
  }
}`

With the current setup we can infer the types from the values, and I don't know of any way to do the opposite

src/models/commands/command.ts
```diff
@@ -1,49 +1,67 @@
...
 import { ArgumentMetadata } from './argumentmetadata';
 
+type toPrim<T = never> = {
+  'string': string
+  'number': number
+  'boolean': boolean
+  'array': T[]
+}
+
+type Metadata = {
```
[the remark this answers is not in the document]
cea2aj: We could actually just use `ArgumentMetadata` here to simplify things

src/models/commands/command.ts
```diff
@@ -1,49 +1,67 @@
...
+export type ArgsForExecute<T extends Record<string, Metadata>> = {
+  [prop in keyof T]: toPrim<toPrim[T[prop]['itemType']]>[T[prop]['type']]
+}
 
+/**
+ * Command interface that contains non static fields and methods
+ * of a Command instance. It requires a type T that defines the
+ * arguments pass to execute()
+ */
+interface CommandExecutable<T> {
```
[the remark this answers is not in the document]
cea2aj: sure, I can move it and simplify

src/models/commands/command.ts
cea2aj: Ya, that type is T

PR #250, opened 2021-09-03
Update ArgumentMetadata to use concrete classes for runtime types

oshi97:
This commit updates the ArgumentMetadata used in custom and built in commands,
so that they now use concrete implementations. This lets us encode additional metadata
like runtime types needed by the yargs instance, without leaking implementation details
to the user.

It also refactors legacy command into one place.

The only caveat with this approach is that you cannot mix and match global and local node_module
usages of jambo, because the metadata classes imported by a user's custom command must come from
the instance of jambo that is running. If need be, we can update jambo to use internal metadata
provided by the concrete implementations, instead of checking for an instanceof a concrete implementation.
Updating jambo in this fashion wouldn't be a breaking change, so I think it should be fine to continue
using instanceof for now.
The nice thing about requiring an instanceof check is we can prevent people from hacking a "working"
argument metadata object because they know implementation details, and force them down a single path.

Additionally, if you use ONLY a global instance of jambo, in order for require('jambo') to work, you must
make sure your node paths are set up properly (using NVM to manage node versions can break this)

J=SLAP-1493
TEST=manual,auto

tested that legacy commands work, and also modern-style commands with static methods, that
still use the old metadata (i.e. what the theme currently uses)

test that importing the various metadata from jambo, and using them in custom commands works.
did this by npm installing jambo from this github branch.

test that v2 and v3 custom command classes get correct validation for static and instance methods


src/commands/LegacyAdapter.ts
```diff
@@ -0,0 +1,46 @@
...
+  return !('prototype' in command);
+}
+/**
+ * Checks whether a custom command class has any arguments that are LegacyArgumentMetadata.
+ */
+function isCommandClassWithLegacyArguments(
+  command: CommandClassWithLegacyArguments | Command<any>
+): command is CommandClassWithLegacyArguments {
+ function isLegacyArgs(args: Record<string, ConcreteArgumentMetadata | LegacyArgumentMetadata>) {
+    for (const metadata of Object.values(args)) {
```
yen-tt, line 36: maybe we can use Object.values(args).some() or every() here?

src/commands/LegacyAdapter.ts
```diff
@@ -0,0 +1,46 @@
...
+import { LegacyArgumentMetadata } from '../models/commands/LegacyArgumentMetadata';
+import adaptCommandWithLegacyArgs, { CommandClassWithLegacyArguments } from './adaptCommandWithLegacyArgs';
+import adaptLegacyCommand, { LegacyCommand } from './adaptLegacyCommand';
+
+type IndeterminateCommand = LegacyCommand | CommandClassWithLegacyArguments | Command<any>;
+
+export default class LegacyAdapter {
+  adapt(command: IndeterminateCommand): Command<any> {
+    const commandClass = isLegacyCommand(command) ? adaptLegacyCommand(command) : command;
+    const commandClassWithModernArgs =
```
yen-tt, line 12: to clarify, our first version of custom command would get wrapped twice in Command instance here right? would it work to use the function `adaptLegacyArguments` in adaptLegacyCommand?

src/models/commands/commandexecutable.ts
```diff
@@ -1,25 +1,34 @@
...
+  StringArrayMetadata,
+  StringMetadata
+} from './concreteargumentmetadata';
 
-/**
- * Represents a record of argument names to their corresponding type
- */
- export type ExecArgumentRecord = {
-  [arg: string]: Primitive | Primitive[]
+type ExecArgs<T extends ArgumentMetadataRecord> = {
```
yen-tt, line 11: ooh nice

tests/fixtures/customcommands/TestCommandWithLegacyArgs.js
```diff
@@ -0,0 +1,32 @@
+const { LegacyArgumentMetadata, LegacyArgumentType } = require('./LegacyArgumentMetadata');
+
+/**
+ * An LegacyCommand implementation for unit tests.
+ * This is an example of our VERY FIRST version of custom commands, before updating all methods but
```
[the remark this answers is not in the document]
oshi97: yes, updated!

src/commands/LegacyAdapter.ts
oshi97: updated!

src/commands/commandimporter.ts
```diff
@@ -80,7 +77,7 @@ export default class CommandImporter {
   /**
    * Validates an imported custom {@link Command} by ensuring the class has all
```
yen-tt, line 78: just a note, since we wrap our legacy custom commands (first and second versions) with the Command class structure, I think this wouldn't catch the missing methods from those commands? they might just throw an error during the adapt/convert to modern command structure process instead when invoking these static and instance methods that might not exist.

src/commands/LegacyAdapter.ts
oshi97: That could work! This just felt like it separated responsibility a little better to me, and would be easier to rip out in the future.

src/commands/commandimporter.ts
oshi97: I moved it into LegacyAdapter! I didn't add additional validation for our oldest version of jambo commands, because we didn't have validation for them before. Tested on the theme's test site to make sure the error message is displayed

src/commands/LegacyAdapter.ts
```diff
@@ -7,8 +9,14 @@ import adaptLegacyCommand, { LegacyCommand } from './adaptLegacyCommand';
 type IndeterminateCommand = LegacyCommand | CommandClassWithLegacyArguments | Command<any>;
 
 export default class LegacyAdapter {
-  adapt(command: IndeterminateCommand): Command<any> {
+  adapt(
+    command: IndeterminateCommand,
+    filePath: string
+  ): Command<any> {
     const commandClass = isLegacyCommand(command) ? adaptLegacyCommand(command) : command;
+    if (!isValidCustomCommand(commandClass)) {
```
yen-tt, line 17: previously, we provide a warning, then ignore this command and continue on with other stuff. Should we keep that format or is throwing an error something we want now?

oshi97: I was thinking it would be nice to force people to fix their custom commands if they were broken. Also right now, if you try to use a malformatted command jambo will say "unknown argument [CommandName]", which I thought could be nice to get rid of.
image
But also kind of out of scope, and nobody other than us are writing custom commands, so it just logs a warning again!

yen-tt (approved):

PR #251, opened 2021-09-07
unify describe with args

oshi97:
This commit updates describecommand to combine information from args
with the return value of describe (a DescribeDefinition) to produce
an instance of DescribeOutput, which is the full output to be used
by UI like sitesgitstorm.

J=SLAP-1493
TEST=manual,auto

test that all jambo card, page, vertical, and build commands work in SGS
test that I can run custom commands locally

### DIFF
--- a/src/commands/describe/describecommand.ts
+++ b/src/commands/describe/describecommand.ts
@@ -51,7 +51,7 @@ const DescribeCommand: Command<any> = class {
     const descriptions = {};
     const commands = this.getCommands();
     const describePromises = commands.map(command => {
-      const recordDescription = (value: DescribeDefinition<ArgumentMetadataRecord>) => {
+      const recordDescription = (value: DescribeDefinition) => {
         descriptions[command.getAlias()] = this.calculateDescribeOutput(command.args(), value);
       }
       const describeValue = command.describe(this._jamboConfig);
@@ -71,7 +71,7 @@ const DescribeCommand: Command<any> = class {
 
   private calculateDescribeOutput(
     args: ArgumentMetadataRecord,
-    describeDefinition: DescribeDefinition<ArgumentMetadataRecord>
+    describeDefinition: DescribeDefinition
   ): DescribeOutput {
     if (!describeDefinition.params) {
       return {
@@ -96,12 +96,10 @@ const DescribeCommand: Command<any> = class {
 }
 
 function isPromise(
-  describeValue: DescribeDefinition<ArgumentMetadataRecord> | Promise<DescribeDefinition<ArgumentMetadataRecord>>
-): describeValue is Promise<DescribeDefinition<ArgumentMetadataRecord>> {
-  if (!('then' in describeValue)) {
-    return false;
-  }
-  return typeof describeValue.then === 'function';
+  describeValue: DescribeDefinition | Promise<DescribeDefinition>
+): describeValue is Promise<DescribeDefinition> {
+  return ('then' in describeValue)
+    && typeof describeValue.then === 'function';
 }
 
 export default DescribeCommand;
--- a/src/commands/describe/describecommand.ts
+++ b/src/commands/describe/describecommand.ts
@@ -1,13 +1,16 @@
-import Command from '../../models/commands/command';
+import Command from '../../models/commands/Command';
+import { ArgumentMetadataRecord } from '../../models/commands/concreteargumentmetadata';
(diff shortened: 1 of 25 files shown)


src/commands/describe/describecommand.ts
```diff
@@ -33,35 +36,72 @@ const DescribeCommand : Command<any> = class {
...
+    const commands = this.getCommands();
+    const describePromises = commands.map(command => {
+      const recordDescription = (value: DescribeDefinition<ArgumentMetadataRecord>) => {
+        descriptions[command.getAlias()] = this.calculateDescribeOutput(command.args(), value);
+      }
+      const describeValue = command.describe(this._jamboConfig);
+      if (!describeValue) {
+        return;
+      }
+      if (isPromise(describeValue)) {
```
yen-tt, line 61: I believe `command.describe(this._jamboConfig);` should no longer by async/returning a promise (themeupgrader command no longer fetch themes on describe (Remove ability of upgrade command to fetch Theme branches. #190)) so I think you can refactor this! might need to update the describe test to non async if not already.

oshi97: I was thinking we could keep the promise handling? It's a little ugly, but I also don't really think it's harming anything (and it seems like something we could want in the future). Technically it would be a breaking change to remove it, though realistically nobody has written a custom command with an async describe so I can remove it if you prefer!

src/models/commands/DescribeDefinition.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export default interface DescribeDefinition<T extends ArgumentMetadataRecord> {
+  displayName: string
+  params?: {
+    [arg in keyof T]: DescribeDefinitionParam<T[arg]>
+  }
+}
+
+export type DescribeDefinitionParam<T extends ConcreteArgumentMetadata> =
+  T extends StringArrayMetadata ? DescribeParamForArray<'string'> :
```
yen-tt, line 24: can we just pass in the type string, number, boolean, instead of the string versions to DescribeParamForArray<T>?

src/commands/describe/describecommand.ts
yen-tt: hm i see, I'm alright with keeping it then!

src/commands/describe/describecommand.ts
```diff
@@ -33,35 +36,72 @@ const DescribeCommand : Command<any> = class {
...
+  }
+}
+
+function isPromise(
+  describeValue: DescribeDefinition<ArgumentMetadataRecord> | Promise<DescribeDefinition<ArgumentMetadataRecord>>
+): describeValue is Promise<DescribeDefinition<ArgumentMetadataRecord>> {
+  if (!('then' in describeValue)) {
+    return false;
+  }
+  return typeof describeValue.then === 'function';
```
yen-tt, line 104: how about `return 'then' in describeValue && typeof describeValue.then === 'function'`?

src/models/commands/DescribeDefinition.ts
oshi97: ah, yeah I think originally I was restricting the type you can specify to one of those strings, but I think that's probably overkill

src/commands/describe/describecommand.ts
oshi97: yeah that's simpler!

src/models/commands/DescribeDefinition.ts
```diff
@@ -13,27 +13,27 @@ import {
  * DescribeOutput provides static type checking for a {@link Command}'s describe(),
  * and ensures that the return value matches the same shape as the {@link Command}'s args.
  */
-export default interface DescribeDefinition<T extends ArgumentMetadataRecord> {
+export default interface DescribeDefinition<T extends ArgumentMetadataRecord = ArgumentMetadataRecord> {
```
yen-tt, line 16: for my own understanding, what's the purpose of the equal part?

oshi97: it sets the default

oshi97: https://mariusschulz.com/blog/generic-parameter-defaults-in-typescript#generic-type-definitions-with-type-parameter-defaults

yen-tt: ah right right

yen-tt (approved):

PR #255, opened 2021-09-16
Export types with API extractor

nmanu1:
Add API extractor to get an API Report, a d.ts bundle, and markdown docs, as in Answers Core. Also, add more exports, documentation, and link to markdown docs in README. Use npm-force-resolutions to upgrade to a validator version that fixes a security vulnerability.

Notes:
- When reviewing, you can ignore the docs files
- Forgotten export warnings in the API Report are out of scope for this PR
- As in Answers Core, there are separate `build-ci` and `api-extractor-ci` scripts, which exit with non-zero exit codes for warning and outdated api-reports
- As in Answers Core, the automatic tests will error out if the docs are out of date

J=SLAP-1495
TEST=manual

Ran the api extractor and inspected the api report, type bundle, and markdown docs. Made changes to the docs and checked that the docs validation script worked as expected.


package.json
```diff
@@ -9,7 +9,11 @@
   ],
   "scripts": {
     "test": "eslint . && jest",
-    "build": "tsc"
+    "build": "tsc && npm run api-extractor && npm run generate-docs",
+    "build-ci": "tsc && npm run api-extractor-ci && npm run generate-docs",
```
oshi97, line 13: what's the difference between  api-extractor and api-extractor-ci?

src/models/commands/DescribeDefinition.ts
```diff
@@ -35,7 +35,7 @@ type DescribeParamForPrimitive<T extends string | number | boolean> = {
   displayName: string
   options?: T[]
   /**
-   * @note 'string', 'number', and 'boolean' are deprecated
+   * @remarks 'string', 'number', and 'boolean' are deprecated
    * These types can be automatically inferred from Command.args()
```
oshi97, line 39: Oh this is my bad but could we specify specify that "these types" means the deprecated types?

src/models/commands/commandexecutable.ts
```diff
@@ -22,13 +27,15 @@ type ExecArgs<T extends ArgumentMetadataRecord> = {
...
+ * @public
  */
 export interface CommandExecutable<T extends ArgumentMetadataRecord> {
   /**
    * Executes the command with the provided arguments.
    *
-   * @param args The arguments passed to the execute command
+   * @param args - The arguments passed to the execute command
    */
   execute(args: ExecArgs<T>): any
```
oshi97, line 40: ah ok again my bad but can we change the return type here to void?

src/models/commands/concreteargumentmetadata.ts
```diff
@@ -32,25 +46,56 @@ interface InternalArrayMetadata {
  * by checking the `instanceof` for an ArgumentMetadata.
  * This allows us to hide the "type" and "itemType" implementation details.
  */
+
+/**
+ * Metadata for an argument with type string.
```
oshi97, line 51: Would `ArgumentMetadata for a string` be slightly simpler?

src/models/commands/concreteargumentmetadata.ts
```diff
@@ -2,11 +2,25 @@ import { ArgumentType, ArgumentMetadata } from './ArgumentMetadata'
 
 /**
  * The base concrete ArgumentMetadata class.
+ *
+ * @public
```
oshi97, line 6: I don't think this class needs to be public, people will only be using the concrete implementation below directly, and not this base class.

src/exports.ts
```diff
@@ -4,14 +4,37 @@ import {
...
+
 
 export {
   StringMetadata,
   StringArrayMetadata,
   BooleanMetadata,
   BooleanArrayMetadata,
   NumberMetadata,
-  NumberArrayMetadata
+  NumberArrayMetadata,
```
oshi97, line 30: maybe we could have a comment or something separating the exported classes from the exported types/interfaces?

README.md
```diff
@@ -8,6 +8,8 @@
 
 Jambo is a JAMStack implementation using Handlebars.
 
+[Full Documentation](./docs/jambo.md)
```
oshi97, line 11: I feel like "Full Documentation" is a little confusing here, since it's just the documentation for our exported types and doesn't document how to use the built in jambo commands, which is below.

src/models/commands/concreteargumentmetadata.ts
oshi97: or maybe `ArgumentMetadata for a string argument`?

oshi97: or maybe something like `ArgumentMetadata for a string argument`?

src/models/commands/concreteargumentmetadata.ts
nmanu1: should I remove it from exports.ts?

README.md
nmanu1: That makes sense. Should I change the name, or just not include it here at all?

src/models/commands/concreteargumentmetadata.ts
oshi97: yup!

README.md
oshi97: let's change the name we'll still want a link to it

src/exports.ts
oshi97: I was just thinking something like the below (nothing fancy)
```ts
export {
  // concrete ArgumentMetadata classes
  StringMetadata,
  StringArrayMetadata,
  BooleanMetadata,
  BooleanArrayMetadata,
  NumberMetadata,
  NumberArrayMetadata
  NumberArrayMetadata,

  // types
  ArgumentMetadata
  //....
```

nmanu1: I can add that

src/models/commands/concreteargumentmetadata.ts
```diff
@@ -2,11 +2,25 @@ import { ArgumentType, ArgumentMetadata } from './ArgumentMetadata'
...
+   */
   isRequired?: boolean
+  /**
+   * Optional, a default value for the argument.
+   */
   defaultValue?: T
+  /**
+   * Optional, a display name for the argument.
+   */
   displayName?: string
```
cea2aj, line 24: I think the ArgumentMetadata should also have an optional displayName

src/models/commands/concreteargumentmetadata.ts
cea2aj: removing it makes sense. We'll just want to make sure that people can easily refer to the ArgumentMetadata interface from the concrete implementation so they can see the full options. Maybe in the remarks for each concrete implementation (e.g. BooleanArgumentMetadata and so on) we include a link to ArgumentMetadata?

nmanu1: actually, I think we can keep it in for completeness of the docs. otherwise the api report gives a forgotten export warning when it's removed

src/models/commands/concreteargumentmetadata.ts
oshi97: Ah actually can we remove this? displayName is only needed by describe, so I wanted to remove it from ArgumentMetadata but missed it here

README.md
nmanu1: like `Documentation for Exported Types`? and would this be better at the bottom of the README as a new section?

src/models/commands/concreteargumentmetadata.ts
nmanu1: Removed!

README.md
oshi97: yeah that makes sense!

etc/jambo.api.md
```diff
@@ -0,0 +1,116 @@
...
+    // (undocumented)
+    readonly type = "boolean";
+}
+
+// @public
+export interface Command<T extends ArgumentMetadataRecord> {
+    // (undocumented)
+    new (...args: any[]): CommandExecutable<T>;
+    args(): T;
+    // Warning: (ae-forgotten-export) The symbol "DescribeDefinition" needs to be exported by the entry point exports.d.ts
```
cea2aj, line 51: Ts there a reason for not exporting DescribeDefinition? Including it would make the docs more complete.

README.md
```diff
@@ -163,3 +163,9 @@ Defaults to false.
 You can signal whether to pass an --isLegacy flag to the theme's upgrade.js.
 This can be useful when you need different/additional upgrade logic to upgrade
 from older versions of a theme. Defaults to false.
+
+### Custom Commands
+
+Custom commands can be created using our exported types.
```
cea2aj, line 169: Can we add an example of creating a built-in command? I think having an example would help people understand the generated docs

src/models/commands/concreteargumentmetadata.ts
cea2aj: Exporting ArgumentMetadataImpl helps document the constructor of the argument metadata classes. If we do leave it in the docs, I suggest we add some remarks to indicate that the class should not be used directly, and that one of the other classes such as BooleanMetadata, NumberMetadata, etc. should be used instead

etc/jambo.api.md
nmanu1: I talked with Tom and Oliver about that, and we're going to add it in a later PR after changing the naming

README.md
cea2aj: edit: adding an example is bit out of scope of exporting the types, so resolving this

scripts/validate-docs.sh
```diff
@@ -4,7 +4,7 @@
 # The script should be ran after new docs are built
 # If there are any git diffs after new docs are generated, the docs are out of date
 
-git diff --exit-code > /dev/null # send stdout to /dev/null to reduce clutter in the CI output
+git diff docs --exit-code > /dev/null # send stdout to /dev/null to reduce clutter in the CI output
```
cea2aj, line 7: After testing this locally, I learned that '--exit-code' needs to go before 'docs' or else the command will break and the script will think the docs look good

package.json
```diff
@@ -13,7 +13,11 @@
     "build-ci": "tsc && npm run api-extractor-ci && npm run generate-docs",
     "api-extractor": "api-extractor run --local --verbose",
     "api-extractor-ci": "api-extractor run --verbose",
-    "generate-docs": "api-documenter markdown --input-folder temp --output-folder docs && rm -rf temp"
+    "generate-docs": "api-documenter markdown --input-folder temp --output-folder docs && rm -rf temp",
+    "preinstall": "npx npm-force-resolutions"
+  },
+  "resolutions": {
+    "validator": "13.6.0"
```
cea2aj, line 20: what was the reason behind forcing this version?

nmanu1: the security/snyk check was failing because previous validator versions had a ReDoS vulnerability. that was fixed in v13.6.0

cea2aj (approved):

PR #260, opened 2021-09-23
Make describe optional

cea2aj:
Make describe optional including for the legacy command interfaces

J=SLAP-1598
TEST=manual, auto

Use the theme test site and comment out the describe method. See that the command no longer shows up in the describe output. Test that when describe is not included, the command still works. 

Add unit test coverage

### DIFF
--- a/src/models/commands/Command.ts
+++ b/src/models/commands/Command.ts
@@ -36,5 +36,5 @@ export default interface Command<T extends ArgumentMetadataRecord> {
    * @returns description of the card command, including paths to
    *          all available cards. returning null causes no description to be output.
    */
-  describe(jamboConfig: JamboConfig): null | DescribeMetadata<T> | Promise<DescribeMetadata<T>>;
+  describe?(jamboConfig: JamboConfig): null | DescribeMetadata<T> | Promise<DescribeMetadata<T>>;
 }
--- a/tests/commands/describe/describecommand.js
+++ b/tests/commands/describe/describecommand.js
@@ -113,3 +113,24 @@ it('deprecated params in a command\'s DescribeMetadata ' +
     }
   });
 })
+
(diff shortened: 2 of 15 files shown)


src/models/commands/Command.ts
```diff
@@ -36,5 +36,5 @@ export default interface Command<T extends ArgumentMetadataRecord> {
    * @returns description of the card command, including paths to
    *          all available cards. returning null causes no description to be output.
    */
-  describe(jamboConfig: JamboConfig): null | DescribeMetadata<T> | Promise<DescribeMetadata<T>>;
+  describe?(jamboConfig: JamboConfig): null | DescribeMetadata<T> | Promise<DescribeMetadata<T>>;
```
yen-tt, line 39: would this be expecting undefined type as well for the return value then? Since the conversion from legacy command to current command structure would have a describe function but internally would return undefined

cea2aj: good point, i'll update

tests/commands/describe/describecommand.js
```diff
@@ -112,4 +112,25 @@ it('deprecated params in a command\'s DescribeMetadata ' +
...
+      }
+    },
+    getAlias() { return 'mocked'; }
+  }
+
+  await new DescribeCommand({}, () => [ mockCommand ]).execute();
+  const descriptions = JSON.parse(consoleSpy.mock.calls[0][0]);
+  consoleSpy.mockClear();
+
+  expect(descriptions.mocked).toBeUndefined();
```
yen-tt, line 135: to double check, DescribeCommand's `getCommandDescriptions()` set `descriptions` field to be `{}` so if there's no command description, then it would console log out '{}' so shouldn't this still be defined?

yen-tt (approved):

PR #278, opened 2021-11-23
add i18next specific locale formatting to support chinese

oshi97:
Prior to this, certain kinds of chinese locales would not be translated properly.
i18next will silently refuse to translate for you if your locale has any underscores in it.

J=TECHOPS-2885
TEST=manual,auto

tested I can do zh-Hant_TW translations in the theme


yen-tt (approved):

yen-tt:
I think SDK translator.js would need to be updated like this too right?

oshi97:
> I think SDK translator.js would need to be updated like this too right?

gratefully for SDK chinese locales we only have zh-Hant and zh-Hans, so it shouldn't be urgent to patch this! I will still put up a PR though (thank you for reminding me!)

PR #279, opened 2021-11-29
Version 1.12.2

oshi97:
* fixes an issue where chinese was not being translated properly (#278).


oshi97:
retested manually with a chinese locale with non i18next formatting (zh_Hant_TW)

tmeyer2115 (approved):

PR #306, opened 2022-04-22
Use reusable workflows

nmanu1:
Update the github workflows to use the callable workflows in the `slapshot-reusable-workflows` repo.

J=SLAP-2005
TEST=auto

- See that the `coverage`, `update_docs`, and `run_tests` workflows run on this PR as expected.
- Test on another branch (`dev/test-reusable-workflows`) and see that docs are updated when needed.
- Fork the repo and commit to master. See that the `sync_develop_and_main` workflow creates a PR as expected.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -9,24 +9,11 @@ on:
   pull_request:
 
 jobs:
-  Coveralls:
-    runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [16.x]
-
-    steps:
-    - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
-      with:
-        node-version: ${{ matrix.node-version }}
-    - run: npm ci
-    - run: git config --global user.email "slapshot@yext.com"
-    - run: git config --global user.name "Jambo run-tests.yml"
-    - run: npx jest --config=jest-coverage.json
-    - name: Coveralls
-      uses: coverallsapp/github-action@master
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
+  call_code_coverage:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/coverage.yml@v1
+    with:
+      test_script: 'git config --global user.email "slapshot@yext.com"
+        && git config --global user.name "Jambo run-tests.yml"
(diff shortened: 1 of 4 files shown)


cea2aj (approved):

.github/workflows/coverage.yml
```diff
@@ -9,24 +9,11 @@ on:
...
-    - run: npx jest --config=jest-coverage.json
-    - name: Coveralls
-      uses: coverallsapp/github-action@master
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+  call_code_coverage:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/coverage.yml@v1
+    with:
+      test_script: 'git config --global user.email "slapshot@yext.com"
```
oshi97, line 15: does specifying `test_script` as a step before `call_code_coverage` also work? I think the `uses` would need to be moved under an additional `steps` key, but not sure if it would work as intended

nmanu1: I don't think that would work because it seems like a reusable workflow has to be called directly from a `job` and can't be called from `steps` (see https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow)

oshi97: ah ok got it

oshi97 (approved):

.github/workflows/coverage.yml
```diff
@@ -9,24 +9,11 @@ on:
...
-    - run: git config --global user.email "slapshot@yext.com"
-    - run: git config --global user.name "Jambo run-tests.yml"
-    - run: npx jest --config=jest-coverage.json
-    - name: Coveralls
-      uses: coverallsapp/github-action@master
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+  call_code_coverage:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/coverage.yml@v1
```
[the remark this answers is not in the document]
nmanu1: Good point. I'll update the `slapshot-reusable-workflows` repo to move these out of the `.github` directory and then update this PR accordingly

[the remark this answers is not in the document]
nmanu1: After looking into it further, reusable workflows are required to be located under `.github/workflows` to be valid, so the prefix will have to remain even though it is a little verbose

PR #308, opened 2022-06-15
improve error message when a page config cannot be parsed

oshi97:
It now always says which file could not be parsed.

J=SLAP-2134
TEST=manual

intentionally make a syntax error in a page config in the theme's test site
see the expected error message

### DIFF
--- a/src/commands/build/sitesgenerator.js
+++ b/src/commands/build/sitesgenerator.js
@@ -21,7 +21,7 @@ const registerCustomHbsHelpers = require('../../handlebars/registercustomhbshelp
 const SystemError = require('../../errors/systemerror');
 const Translator = require('../../i18n/translator/translator');
 const UserError = require('../../errors/usererror');
-const { info } = require('../../utils/logger');
+const { info, error } = require('../../utils/logger');
 
 class SitesGenerator {
   constructor(jamboConfig) {
@@ -59,10 +59,12 @@ class SitesGenerator {
             true
           );
         } catch (err) {
-          if (err instanceof SyntaxError) {
+          if (err instanceof SyntaxError ||
+            err.message && err.message.includes('Invalid regular expression')) {
             throw new UserError(
-              `JSON SyntaxError: could not parse file ${path}`, err.stack);
+              `Could not parse JSON page config at ${path}`, err.stack);
           } else {
+            error('Error parsing JSON page config at', path);
             throw err;
           }
         }


src/commands/build/sitesgenerator.js
```diff
@@ -59,10 +59,12 @@ class SitesGenerator {
...
           );
         } catch (err) {
-          if (err instanceof SyntaxError) {
+          if (err instanceof SyntaxError ||
+            err.message && err.message.includes('Invalid regular expression')) {
             throw new UserError(
-              `JSON SyntaxError: could not parse file ${path}`, err.stack);
+              `Could not parse JSON page config at ${path}`, err.stack);
           } else {
+            error('Error parsing JSON page config at', path);
```
yen-tt, line 67: can we not print err.stack here as well?

src/commands/build/sitesgenerator.js
```diff
@@ -59,10 +59,12 @@ class SitesGenerator {
             true
           );
         } catch (err) {
-          if (err instanceof SyntaxError) {
+          if (err instanceof SyntaxError ||
+            err.message && err.message.includes('Invalid regular expression')) {
             throw new UserError(
-              `JSON SyntaxError: could not parse file ${path}`, err.stack);
+              `Could not parse JSON page config at ${path}`, err.stack);
```
yen-tt, line 65: can we update the message to specify it's a syntax error? Otherwise this one and the error in else is basically the same so we could just group it

oshi97: updated!

src/commands/build/sitesgenerator.js
oshi97: The err stack gets printed afterwards. I didn't like this though, this first error message would get separated from the later one. I added the UnknownError class so we could insert additional messages around errors we didn't know whose fault was. Something that's always bothered me about jambo errors is that we force them to be either SystemError or UserError, but most errors we aren't able to accurately pin blame on. We can also pin blame on the User or System incorrectly.

oshi97: For instance in buildcommand.js before if we didn't recognize an error as either System or User we just... assumed it was the user's fault... which doesn't seem right.

yen-tt: gotcha, yea having an UnknownError type makes sense

PR #330, opened 2024-01-31
Resolve Vulnerabilities

EmilyZhang777:
This PR resolves the following vulnerabilities

Bump `tmpl`'s version to 1.0.5 to avoid uncontrolled resource consumption when formatting a string
J=VULN-38389
TEST=auto

Ran npm run test. Also made sure running jambo commands locally works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@babel/plugin-transform-object-assign": "^7.12.1",
     "@types/jest": "^26.0.23",
     "eslint": "^7.8.1",
-    "jest": "^25.4.0",
+    "jest": "^29.7.0",
     "prettier": "^2.2.1",
     "shell-quote": "^1.7.2"
   },
--- a/package.json
+++ b/package.json
@@ -45,9 +45,12 @@
     "@babel/plugin-transform-object-assign": "^7.12.1",
     "@types/jest": "^26.0.23",
     "eslint": "^7.8.1",
-    "jest": "^29.7.0",
+    "jest": "^25.4.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -49,6 +49,9 @@
     "prettier": "^2.2.1",
     "shell-quote": "^1.7.2"
   },
+  "overrides": {
+    "tmpl": "^1.0.5"
```
nmanu1, line 53: same comment as the other PR about avoiding overrides unless necessary

nmanu1 (commented):
the PR description can be updated now that we're only updating `tmpl` and not jest. I think it's helpful for the description to include exactly which versions are vuln free (e.g. something like `tmpl@1.0.5+ here). that way we have it for posterity

package.json
EmilyZhang777: updated

EmilyZhang777:
Updated!

src/utils/customcommands/commandexecuter.js
```diff
@@ -28,28 +28,28 @@ exports.CustomCommandExecuter = class {
             command.getArgs(),
             {
                 cwd: command.getCwd(),
-                shell: true,
+                shell: false,
```
nmanu1, line 31: what is the affect of this change? and have you tried executing any jambo commands to make sure they still work?

EmilyZhang777: Tested locally with slapshot-v1.28.1-test and ran a few jambo commands

